PR #27, opened 2018-10-16
Enhance versioning and building process

### LINKED ISSUE #15: Add Travis Build

adrianmo:
- Create Makefile with options to build, create Docker image, run tests, and format code.
- Update Travis build process (fix #15)
- Obtain version number at build time based on the git tag. If there are commits on top of the tagged commit, the number of additional commits will be appended to the tag (`0.1.0-4` if there are four commits on top of the `0.1.0` tag). If the working tree has local modifications, the `-dirty` suffix will be added.
- Show Git SHA when displaying the operator version.
- Create Dockerfile to build the operator Docker image.
- Create utility scrips under `scripts` to check that all Go files include the license header and comply with `gofmt` code style.

Signed-off-by: Adrian Moreno <adrian@morenomartinez.com>


Dockerfile
```diff
@@ -0,0 +1,30 @@
+FROM golang:1.10.1-alpine3.7 as go-builder
```
arvindkandhare: Why are we using the `as` operator here? I think change of the base image should be an intrusive change.

Dockerfile
```diff
@@ -0,0 +1,30 @@
...
+
+ARG PROJECT_NAME=zookeeper-operator
+ARG REPO_PATH=github.com/pravega/$PROJECT_NAME
+ARG BUILD_PATH=${REPO_PATH}/cmd/${PROJECT_NAME}
+
+# Build version and commit should be passed in when performing docker build
+ARG VERSION=0.0.0-localdev
+ARG GIT_SHA=0000000
+
+COPY pkg /go/src/${REPO_PATH}/pkg
```
arvindkandhare: setting up `GO_PATH` and `GO_HOME` and then running `go install` may be the correct way of doing it. Is there any specific reason we are copying the bits by hand?

Dockerfile
```diff
@@ -0,0 +1,30 @@
...
+COPY cmd /go/src/${REPO_PATH}/cmd
+COPY vendor /go/src/${REPO_PATH}/vendor
+
+RUN GOOS=linux GOARCH=amd64 CGO_ENABLED=0 go build -o ${GOBIN}/${PROJECT_NAME} \
+    -ldflags "-X ${REPO_PATH}/pkg/version.Version=${VERSION} -X ${REPO_PATH}/pkg/version.GitSHA=${GIT_SHA}" \
+    $BUILD_PATH
+
+# =============================================================================
+FROM alpine:3.7 AS final
```
arvindkandhare: Can you explain why we have two `FROM` here. I do not understand the problem we are solving/how it works. Can you explain in the PR comment/comment in the file what we are trying to do here?

Makefile
```diff
@@ -0,0 +1,52 @@
...
+	-o bin/$(PROJECT_NAME) cmd/$(PROJECT_NAME)/main.go
+
+build-image:
+	docker build --build-arg VERSION=$(VERSION) --build-arg GIT_SHA=$(GIT_SHA) -t $(REPO):$(VERSION) .
+	docker tag $(REPO):$(VERSION) $(REPO):latest
+
+test:
+	go test $$(go list ./... | grep -v /vendor/)
+
+login:
```
arvindkandhare: Who provides these env vars during travis build?

scripts/check_license.sh
```diff
@@ -0,0 +1,17 @@
+#!/usr/bin/env bash
+# exit immediately when a command fails
+set -e
+# only exit with zero if all commands of the pipeline exit successfully
+set -o pipefail
+# error on unset variables
+set -u
+
+licRes=$(
```
arvindkandhare: I think the correct way of doing it is to use one of the existing go tools instead of rolling our own custom checks. Ideally it should read from a `license header` file and match header 1:1. Right now looks like we are looking at only a subset of a pattern.

Dockerfile
adrianmo: It's called a multi-stage build. The idea behind is that we use the first stage to configure and build our application, including all the necessary tooling and dependencies. Then, the second stage normally uses a lightweight base image (e.g. `alpine`) and copies the binary from the first stage, keeping the final image clean and light. You can read more about it here: https://docs.docker.com/develop/develop-images/multistage-build/

Dockerfile
adrianmo: The `as` is just to name stages. In this case, the first stage is called `go-builder`. We can use that reference to obtain files from a different stage.

Dockerfile
adrianmo: These are the only bits that we need to build the operator. We could copy the entire root directory to reduce the number of lines from 3 to 1, but we would be copying useless data.

Makefile
adrianmo: These Docker variables are configured in the Travis job, so they are present during the build process.

scripts/check_license.sh
adrianmo: You are right. We should be checking the entire license header text. I've created the following issue to deal with it in the future: https://github.com/pravega/zookeeper-operator/issues/29

Dockerfile
arvindkandhare: Thanks for the explanation.

Dockerfile
arvindkandhare: The `go` cmd `install` (as described here https://nanxiao.gitbooks.io/golang-101-hacks/content/posts/go-build-vs-go-install.html) does exactly the same thing you are trying to accomplish. My question was about the need to do it ourselves.

Dockerfile
arvindkandhare: Thanks.

scripts/check_license.sh
arvindkandhare: This comment is twofold. First part is using a standard go tool and second is checking the whole license file. The issue just covers the second part.

adrianmo: @arvindkandhare what standard go tool are you referring to?

Dockerfile
adrianmo: @arvindkandhare I still don't understand how we can benefit from `go install`. AFAIK, `go install` builds the project and moves the binary to a specific location, but we don't need to move the binary anywhere else since we will copy it to the final destination in the next stage. Can you please elaborate a bit more your suggestion? Or we can also discuss it offline.

Makefile
```diff
@@ -0,0 +1,52 @@
...
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+
+SHELL=/bin/bash -o pipefail
+
+PROJECT_NAME=zookeeper-operator
+REPO=pravega/$(PROJECT_NAME)
+VERSION=$(shell git describe --always --tags --dirty)
+GIT_SHA=$(shell git rev-parse --short HEAD)
```
maddisondavid: Is this required?  The `VERSION` from `git describe` has the HEAD commit SHA at the end

scripts/check_license.sh
maddisondavid: I agree on both counts (checking entire license and using a tool for it) but I also don't think this should stop this being merged and handled separately and more extensively in #29

Makefile
adrianmo: Actually, `VERSION` will contain the commit SHA if there's no tag available. Otherwise it'll refer to the closest tag. After the first release, there will always be a tag to be referenced. 

However, I now realized that the commit SHA is also added when new commits are added on top of the tag. In the last command, I would like to see `0.0.0-1` instead of `0.0.0-1-g8c44b62`. 

@maddisondavid any ideas?

```
$ git describe --always --tags --dirty
876567a
$ git tag 0.0.0
$ git describe --always --tags --dirty
0.0.0
$ touch foo
$ git add foo
$ git describe --always --tags --dirty
0.0.0-dirty
$ git commit -m "Add foo"
$ git describe --always --tags --dirty
0.0.0-1-g8c44b62
```

adrianmo: I've fixed the above issue in the last commit.

PR #35, opened 2018-10-23
Fixup AccessMode in example/cr-detailed.yaml

jdmaguire:
Mundane change.

I was using example/cr-detailed.yaml and it did not work. Found that the issue was that the sample value for spec.persistence.accessModes is incorrect.


spiegela:
Looks good. Thanks!

PR #60, opened 2019-04-08
Update to support dynamic zookeeper naming

tmarcu:
This change enables the cluster to use the zookeeper-client directly
instead of having to duplicate the service.

Signed-off-by: Tudor Marcu <Tudor.Marcu@emc.com>


EronWright:
Could you explain the underlying issue and motivation for this?  Thanks.

tmarcu:
I've worked through it a bit more, and decided is the wrong solution. These are actually easy to update in my controller's reconcile loop for zookeeper, so I think this operator should be left alone 👍

PR #64, opened 2019-05-02
Issue 63: Enabling ephemeral storage for zookeeper.

### LINKED ISSUE #63: Allow Zookeepers to be run with ephemeral storage.
Currently the operator only allows for persistent volumes, however for certain use cases it can be very useful to have ephemeral storage instead of persistent storage.

This could be implemented like the etcd-operator, where `EmptyDir` is used when no PVC spec is provided. The issue there is that the change would be backwards incompatible. **The API documentation actually states that this is the behavior, but it isn't implemented.**

The changes would be relatively minimal and would just require edits to the statefulSet generator and api.

HoustonPutman:
### Changelog Description

This PR allows users to create zookeepers with ephemeral storage instead of PVCs.

To enable ephemeral storage, omit 
```
spec:
    persistence
````

This change breaks backwards compatibility, but it does adhere to the documentation of the Zookeeper CRD: https://github.com/pravega/zookeeper-operator/blob/68336c7f9fcd51437ec1cfd3fcac40d172610938/pkg/apis/zookeeper/v1beta1/zookeepercluster_types.go#L68-L71

### Purpose of the change

Fix #63 

### How to verify

Start the operator, create a Zk cluster without persistence. 

Signed-off-by: Houston Putman <hputman1@bloomberg.net>

### DIFF
--- a/pkg/controller/zookeepercluster/zookeepercluster_controller.go
+++ b/pkg/controller/zookeepercluster/zookeepercluster_controller.go
@@ -467,7 +467,7 @@ func (r *ReconcileZookeeperCluster) yamlConfigMap(instance *zookeeperv1beta1.Zoo
 }
 
 func (r *ReconcileZookeeperCluster) reconcileFinalizers(instance *zookeeperv1beta1.ZookeeperCluster) (err error) {
-	if instance.Spec.Persistence.VolumeReclaimPolicy != zookeeperv1beta1.VolumeReclaimPolicyDelete {
+	if instance.Spec.Persistence != nil && instance.Spec.Persistence.VolumeReclaimPolicy != zookeeperv1beta1.VolumeReclaimPolicyDelete {
 		return nil
 	}
 	if instance.DeletionTimestamp.IsZero() {
--- a/pkg/zk/generators.go
+++ b/pkg/zk/generators.go
@@ -146,13 +146,13 @@ func makeZkPodSpec(z *v1beta1.ZookeeperCluster, volumes []v1.Volume) v1.PodSpec 
 		Containers: []v1.Container{zkContainer},
 		Affinity:   z.Spec.Pod.Affinity,
 		Volumes:    volumes,
-		TerminationGracePeriodSeconds: &z.Spec.Pod.TerminationGracePeriodSeconds,
 	}
 	if reflect.DeepEqual(v1.PodSecurityContext{}, z.Spec.Pod.SecurityContext) {
 		podSpec.SecurityContext = z.Spec.Pod.SecurityContext
 	}
 	podSpec.NodeSelector = z.Spec.Pod.NodeSelector
 	podSpec.Tolerations = z.Spec.Pod.Tolerations
+	podSpec.TerminationGracePeriodSeconds = &z.Spec.Pod.TerminationGracePeriodSeconds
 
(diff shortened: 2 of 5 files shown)

### CODE REVIEW

**Comment by HoustonPutman** [comment]
Rebased to comply with new `Persistence` option.

Now ephemeral storage is enabled if `Persistence` is not provided.

**Comment by Tristan1900** [comment]
I agree with you @adrianmo, this is a good change since it just implements what the doc promises. I would vote to make the change.

**Comment by HoustonPutman** [comment]
I should also mention, that I updated the format checking script to output the diffs of the expected vs. actual formatting. Making it much easier to debug, since the formatting is tested against multiple go versions.

**Comment by HoustonPutman** [comment]
Have y'all had a chance to look at the implementation?

**Comment by prune998** [comment]
Any issue on that ? Can it be merged ?

File: pkg/controller/zookeepercluster/zookeepercluster_controller.go [hunk 1]
```diff
@@ -467,7 +467,7 @@ func (r *ReconcileZookeeperCluster) yamlConfigMap(instance *zookeeperv1beta1.Zoo
 }
 
 func (r *ReconcileZookeeperCluster) reconcileFinalizers(instance *zookeeperv1beta1.ZookeeperCluster) (err error) {
```
spiegela: Looks like we need coverage of `reconcileFinalizers`, but it isn't currently there, so I'll file that in a separate issue.

HoustonPutman:
Thanks for approving @spiegela! Can we get this merged?

spiegela:
@pbelgundi can you also provide a review for this?

priyavj08:
when can we expect this to be merged?

thanks

pkg/zk/generators.go
```diff
@@ -32,8 +32,30 @@ func headlessSvcName(z *v1beta1.ZookeeperCluster) string {
...
+	if persistence != nil {
+		pvcs = append(pvcs, v1.PersistentVolumeClaim{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:   zkDataVolume,
+				Labels: map[string]string{"app": z.GetName()},
+			},
+			Spec: persistence.PersistentVolumeClaimSpec,
+		})
+	} else {
+		extraVolumes = append(extraVolumes, v1.Volume{
```
pbelgundi, line 51: Our zookeeperStartup.sh script relies on **persistent volumes** to be able to handle server restarts which will be broken when using ephemeral volumes.

spiegela:
> Defaulting to ephemeral storage breaks our DR Strategy. While I like the idea of using ephemeral storage with Zookeeper, at this point it can't be made a default option.

@pbelgundi If we make this a non-default option, is everything else in this PR kosher?  If so, perhaps we should create a parallel PR with the changes we need to get this unblocked.

pbelgundi:
> > Defaulting to ephemeral storage breaks our DR Strategy. While I like the idea of using ephemeral storage with Zookeeper, at this point it can't be made a default option.
> 
> @pbelgundi If we make this a non-default option, is everything else in this PR kosher? If so, perhaps we should create a parallel PR with the changes we need to get this unblocked.

Yes, I'm good as long as existing defaults don't change.

HoustonPutman:
Better implementation available at #215

PR #67, opened 2019-06-12
Issue 66: Make zoo.cfg writable

### LINKED ISSUE #66: Wrong Cluster Membership in Dynamic Config File
**Problem description**
We have suspicions that Pravega components (e.g., BK, SSS) are having trouble to interact with Zookeeper sometimes, which leads to failures. In this sense, an hypothesis we have is that the configuration of Zookeeper may be behind those problems. In a simple experiment, we deployed 3 Zookeeper instances with the Zookeeper Operator, and in the dynamic config file of each instance, we see the following:

(zookeeper-0)
```
/ # cat /data/zoo.cfg.dynamic
server.1=zookeeper-0.zookeeper-headless.default.svc.cluster.local:2888:3888:participant;2181
```
(zookeeper-1)
```
/ # cat /data/zoo.cfg.dynamic
server.1=zookeeper-0.zookeeper-headless.default.svc.cluster.local:2888:3888:participant;0.0.0.0:2181
server.2=zookeeper-1.zookeeper-headless.default.svc.cluster.local:2888:3888:observer;0.0.0.0:2181
```
(zookeeper-2)
```
/ # cat /data/zoo.cfg.dynamic
server.1=zookeeper-0.zookeeper-headless.default.svc.cluster.local:2888:3888:participant;0.0.0.0:2181
server.2=zookeeper-1.zookeeper-headless.default.svc.cluster.local:2888:3888:participant;0.0.0.0:2181
server.3=zookeeper-2.zookeeper-headless.default.svc.cluster.local:2888:3888:observer;0.0.0.0:2181
```
So, apparently every Zookeeper instance has a different view of the cluster, which may lead to problems. According to the documentation of Zookeeper, all the instances should have a consistent configuration.

**Problem location**
Zookeeper operator dynamic config generation.

**Suggestions for an improvement**
At first glance, all the instances should have the same membership in `zoo.cfg.dynamic` with all the members of the Zookeeper cluster. Also, we may need to verify whether the role setting (observer/participant) is correctly set in this context.

Tristan1900:
### Changelog Description
This PR makes the zoo.cfg writable

### Purpose of the change
Fix #66

### How to verify
Start 2 nodes and log into them, check the `zoo.cfg` file in the `/data/conf` dir to see if it contains the latest  `zoo.cfg.dynamic` file and check  `zoo.cfg.dynamic` to see if they have the latest info. 

Signed-off-by: wenqimou <452787782@qq.com>

### DIFF
--- a/docker/bin/zookeeperStart.sh
+++ b/docker/bin/zookeeperStart.sh
@@ -15,10 +15,11 @@ source /conf/env.sh
 source /usr/local/bin/zookeeperFunctions.sh
 
 HOST=`hostname -s`
-DATA_DIR=/data
+ZKROOT=/zookeeper
+DATA_DIR=$ZKROOT/data
 MYID_FILE=$DATA_DIR/myid
 LOG4J_CONF=/conf/log4j-quiet.properties
-ZOOCFGDIR=$DATA_DIR/conf
+ZOOCFGDIR=$ZKROOT/conf
 DYNCONFIG=$DATA_DIR/zoo.cfg.dynamic
 
 # Copy over the zoo.cfg file in the configmap to make it writable
--- a/docker/bin/zookeeperStart.sh
+++ b/docker/bin/zookeeperStart.sh
@@ -20,6 +20,10 @@ MYID_FILE=$DATA_DIR/myid
 LOG4J_CONF=/conf/log4j-quiet.properties
 DYNCONFIG=$DATA_DIR/zoo.cfg.dynamic
 
+# Copy over the zoo.cfg file in the configmap to make it writable
+ZOOCFGDIR=$DATA_DIR
+cp /conf/zoo.cfg /data/zoo.cfg
+
 # Extract resource name and this members ordinal value from pod hostname
 if [[ $HOST =~ (.*)-([0-9]+)$ ]]; then
     NAME=${BASH_REMATCH[1]}
--- a/docker/bin/zookeeperStart.sh
+++ b/docker/bin/zookeeperStart.sh
@@ -18,11 +18,16 @@ HOST=`hostname -s`
 DATA_DIR=/data
 MYID_FILE=$DATA_DIR/myid
 LOG4J_CONF=/conf/log4j-quiet.properties
+ZOOCFGDIR=$DATA_DIR/conf
 DYNCONFIG=$DATA_DIR/zoo.cfg.dynamic
 
 # Copy over the zoo.cfg file in the configmap to make it writable
-ZOOCFGDIR=$DATA_DIR
-cp /conf/zoo.cfg /data/zoo.cfg
+if [ -d $ZOOCFGDIR ]; then
+  rm -r $ZOOCFGDIR
+fi
+mkdir $ZOOCFGDIR
+cp /conf/zoo.cfg $ZOOCFGDIR
+cp /conf/log4j.properties $ZOOCFGDIR
 
 # Extract resource name and this members ordinal value from pod hostname
 if [[ $HOST =~ (.*)-([0-9]+)$ ]]; then
--- a/docker/bin/zookeeperReady.sh
+++ b/docker/bin/zookeeperReady.sh
@@ -15,7 +15,8 @@ source /conf/env.sh
 source /usr/local/bin/zookeeperFunctions.sh
 
 HOST=`hostname -s`
-DATA_DIR=/data
+ZKROOT=/zookeeper
+DATA_DIR=$ZKROOT/data
 MYID_FILE=$DATA_DIR/myid
 LOG4J_CONF=/conf/log4j-quiet.properties
 
--- a/docker/bin/zookeeperTeardown.sh
+++ b/docker/bin/zookeeperTeardown.sh
@@ -14,7 +14,8 @@ set -ex
 source /conf/env.sh
 source /usr/local/bin/zookeeperFunctions.sh
 
-DATA_DIR=/data
+ZKROOT=/zookeeper
+DATA_DIR=$ZKROOT/data
 MYID_FILE=$DATA_DIR/myid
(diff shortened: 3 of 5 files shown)


Tristan1900:
One problem is that log4j seems not working. Checking what happened.
Also, do we need to clean up old  `zoo.cfg.dynamic` over time? It is accumulating.

docker/bin/zookeeperStart.sh
```diff
@@ -20,6 +20,10 @@ MYID_FILE=$DATA_DIR/myid
 LOG4J_CONF=/conf/log4j-quiet.properties
 DYNCONFIG=$DATA_DIR/zoo.cfg.dynamic
 
+# Copy over the zoo.cfg file in the configmap to make it writable
+ZOOCFGDIR=$DATA_DIR
```
RaulGracia: Is it safe to set this environment variable here or could it be overwritten in some situations? Moreover, doing a `printenv` in a Zookeeper pod I see: `ZOO_CONF_DIR=/conf` and `ZOOCFGDIR=/conf`. Do we need to overwrite both?

docker/bin/zookeeperStart.sh
```diff
@@ -20,6 +20,10 @@ MYID_FILE=$DATA_DIR/myid
...
+cp /conf/zoo.cfg /data/zoo.cfg
```
fpj: I don't have any major concern with this change, but for organization purposes, we may want to have the configuration files in a path that ends with `conf/`. For example, we could have:

1-  a common root `zookeeper`
2- the data dir being `zookeeper/data`
3- the conf dir being `zookeeper/conf`

Tristan1900: Sounds reasonable. I put the config files into the '/conf' in the '/data' for now. Since `/data` is the mount point of a pvc, changing the name will require a new release for the operator. Right now the PR only requires the release for zookeeper image.

Tristan1900:
@spiegela Hi Aaron, It would be great if you could take a look at the change when you have time, your input will be really helpful. Thanks!

docker/bin/zookeeperStart.sh
fpj: If you agree that this is a change we want to do in the future, then do you want to create an issue to track it?

RaulGracia:
I'm inspecting the behavior of the Zookeeper cluster with the changes of this branch (using the image that @Tristan1900 posted at `tristan1900/zookeeper:0.2`):
1) The environment variables are not changed for all the users, as executing a `printenv` we have:
```
ZOOCFGDIR=/conf
ZOO_CONF_DIR=/conf
```
Maybe only the process that executes the script sees the proper variables set.
Also, there are other variables not related but that look conflicting:
```
ZOO_STANDALONE_ENABLED=true (env var)
standaloneEnabled=false (in zoo.cfg) <- this is the proper setting
```
2) I do see the configuration folder copied in the `/data` volume and the dynamic config files being created:
```
bash-4.4# pwd
/data/conf
bash-4.4# ls
log4j.properties           zoo.cfg.dynamic.200000000  zoo.cfg.dynamic.200000008  zoo.cfg.dynamic.200000012  zoo.cfg.dynamic.20000001d
zoo.cfg                    zoo.cfg.dynamic.200000003  zoo.cfg.dynamic.20000000a  zoo.cfg.dynamic.20000001a
```
3) The static `zoo.cfg` file is also updated with the latest dynamic configuration:
```
bash-4.4# cat zoo.cfg
initLimit=10
syncLimit=2
skipACL=yes
4lw.commands.whitelist=cons, envi, conf, crst, srvr, stat, mntr, ruok
tickTime=2000
dataDir=/data
reconfigEnabled=true
standaloneEnabled=false
dynamicConfigFile=/data/conf/zoo.cfg.dynamic.20000001d
```

If we could fix the environment variable settings for all the users, this would be sufficient to solve the issue. 

As an alternative, @pbelgundi also proposes to analyze whether the config map in `/conf` is being used at all, and if it isn't, then we could just make `/conf` a writable location.

Both options are fine, but I would be slightly inclined for the second one, as deleting the config map from `/conf` may solve the issue and simplify the operator at the same time.

docker/bin/zookeeperStart.sh
pbelgundi: I think this needs to be changed in zkEnv.sh.

docker/bin/zookeeperStart.sh
pbelgundi: I don't think moving configuration to /data directory is a good solve to this problem. At some point if we need to purge data folder , this will create issues. It could be a reasonable temporary workaround  though. Configurations should be maintained under a separate 'conf' folder, which could be "/conf" or "/zookeeper/conf"
Please see my other updates to Issue #66.

pbelgundi:
As zookeeper is at the center of Pravega operation, lets make sure we test this change thoroughly before pushing it. It may be a good idea to run system tests on zk with these changes,to make sure we don't introduce any new issues.

RaulGracia:
@pbelgundi you are right, but in system tests we only use 1 instance of Zookeeper, so the change is not likely to be relevant. Instead, it would be better to ask @sumit-bm to reproduce few open issues that may benefit from this change (e.g., forced failover scenarios, longevity runs).

docker/bin/zookeeperStart.sh
Tristan1900: Got it. Changed to the corresponding structure. Please review it.

Tristan1900:
@RaulGracia Thanks for the confirmation. Yes, please use `tristan1900/zookeeper:0.2` when testing, the current PR is also changing the operator code.

RaulGracia:
@Tristan1900 is this PR necessary once #69 has been merged? If not, please could you close it?

Tristan1900:
@RaulGracia Sure!

PR #69, opened 2019-06-19
Issue 66: Incorrect cluster membership

### LINKED ISSUE #66: Wrong Cluster Membership in Dynamic Config File

pbelgundi:
### Change log description
Zookeeper dynamic configuration parameters (ex: servers in cluster) can change at run-time, and when that happens, the changed configuration needs to be persisted to disk for future reference. However, in the current zookeeper cluster, "/conf" directory is read-only (source for a config-map), due to which new dynamic configurations could not be persisted, giving rise to inconsistencies in cluster state. 
This PR also fixes the following issues with Readiness Probe :
1. Always failing on servers that start as "observers" because of `exit 1` added in block that upgrades observers to participants.
2. Post deletion and restart Server-0 remains as "observer" and never changes to "participant" as this server was ignored by the readiness probe script.
3. Added 'nslookup' to make sure upgrade is not attempted when service is not yet available ( for the very 1st probe on server 0, that has not yet started)

### Purpose of the change
Fixes #66

### What the code does
This fix changes the configuration location for zookeeper from "/conf" to "/data/conf" which is a writable folder, so new dynamic configurations can get persisted. All necessary files, including static configuration file "zoo.cfg" and copied into the new conf location, before startup.
Updated the 'zk-operator' image repository location to correctly point to "pravega/zookeeper-operator". 

### How to verify it
After this fix, did not see this error in zk logs: "Error writing next dynamic config file to disk".
Also cluster was in a consistent state, with exactly one node being "leader" and other being "followers".
Checked by executing "zkCli.sh status" on each server.

Also saw multiple dynamic configuration files being, as new severs were added to the cluster. The last one has the correct complete configuration and "/data/conf/zoo.cfg" points to it : 
```
/zookeeper-3.5.4-beta # ls -lt /data/conf/
-rw-r--r--    1 root     root           235 Jun 19 03:01 zoo.cfg
-rw-r--r--    1 root     root           484 Jun 19 03:01 zoo.cfg.dynamic.100000065
-rw-r--r--    1 root     root           387 Jun 19 03:01 zoo.cfg.dynamic.100000062
-rw-r--r--    1 root     root           481 Jun 19 03:00 zoo.cfg.dynamic.100000052
-rw-r--r--    1 root     root           387 Jun 19 03:00 zoo.cfg.dynamic.100000039
-rw-r--r--    1 root     root           290 Jun 19 03:00 zoo.cfg.dynamic.100000036
-rw-r--r--    1 root     root           384 Jun 19 03:00 zoo.cfg.dynamic.10000002c

/zookeeper-3.5.4-beta # cat /data/conf/zoo.cfg
initLimit=10
. . .
dynamicConfigFile=/data/conf/zoo.cfg.dynamic.100000065
```

On pod deletion ( 2 at a time) and restart, again new dynamic configuration files get generated on all servers, showing the change in cluster state:
```
/zookeeper-3.5.4-beta # ls -lt /data/conf/
total 56
-rw-r--r--    1 root     root           235 Jun 19 03:05 zoo.cfg
-rw-r--r--    1 root     root           484 Jun 19 03:05 zoo.cfg.dynamic.10000012a
-rw-r--r--    1 root     root           387 Jun 19 03:05 zoo.cfg.dynamic.100000127
-rw-r--r--    1 root     root           481 Jun 19 03:04 zoo.cfg.dynamic.100000111
-rw-r--r--    1 root     root           387 Jun 19 03:03 zoo.cfg.dynamic.1000000e5
-rw-r--r--    1 root     root           484 Jun 19 03:01 zoo.cfg.dynamic.100000065
-rw-r--r--    1 root     root           387 Jun 19 03:01 zoo.cfg.dynamic.100000062
-rw-r--r--    1 root     root           481 Jun 19 03:00 zoo.cfg.dynamic.100000052
-rw-r--r--    1 root     root           387 Jun 19 03:00 zoo.cfg.dynamic.100000039
-rw-r--r--    1 root     root           290 Jun 19 03:00 zoo.cfg.dynamic.100000036
-rw-r--r--    1 root     root           384 Jun 19 03:00 zoo.cfg.dynamic.10000002c

/zookeeper-3.5.4-beta # cat /data/conf/zoo.cfg
initLimit=10
. . .
dynamicConfigFile=/data/conf/zoo.cfg.dynamic.10000012a
```

All unit tests executed successfully.

Tested this fix on GKE as well as PKS clusters.
Deleted servers restart successfully and join the zk ensemble.
Server-0 which earlier used to stay in "obsever" state post restart, now  properly  migrates to "participant" after 1st readiness probe.
No readiness probe failures for servers starting as "observers"


Gopkg.lock
```diff
@@ -761,6 +761,7 @@
     "pkg/version",
     "plugin/pkg/client/auth/exec",
     "plugin/pkg/client/auth/gcp",
+    "plugin/pkg/client/auth/oidc",
```
RaulGracia: Is necessary for the issue being fixed?

docker/bin/zookeeperStart.sh
```diff
@@ -99,9 +99,32 @@ if [[ "$WRITE_CONFIGURATION" == true ]]; then
...
+then
+echo "Creating writable conf directory."
+mkdir $ZOOCFGDIR
+cp -f /conf/zoo.cfg $ZOOCFGDIR
+cp -f /conf/log4j.properties $ZOOCFGDIR
+cp -f /conf/log4j-quiet.properties $ZOOCFGDIR
+cp -f /conf/env.sh $ZOOCFGDIR
+fi
+
+if [[ "$WRITE_CONFIGURATION" == false &&  "$REGISTER_NODE" == false ]]; then
```
RaulGracia: Minor: extra space after `&&`.

docker/bin/zookeeperStart.sh
```diff
@@ -99,9 +99,32 @@ if [[ "$WRITE_CONFIGURATION" == true ]]; then
...
+  # We get here only on server restart...
+  echo Printing bootstrap dynamic config file
```
RaulGracia: Minor: We are using `echo` with and without `"`. It would be better to consolidate.

Gopkg.lock
pbelgundi: This is a generated file, but not sure why it is checked-in...

docker/bin/zookeeperStart.sh
pbelgundi: ok

docker/bin/zookeeperStart.sh
pbelgundi: ok

docker/bin/zookeeperStart.sh
```diff
@@ -99,9 +98,35 @@ if [[ "$WRITE_CONFIGURATION" == true ]]; then
...
+  cat $DYNCONFIG
+  echo Printing Static Config
+  STATIC_CONFIG=`cat $ZOOCFGDIR/zoo.cfg`
+  echo Writing configuration gleaned from zookeeper ensemble
+  echo "$CONFIG" | grep -v "^version="> $DYNCONFIG
+  cat $DYNCONFIG
+  #Setting dynamicConfigFile=/data/zoo.cfg.dynamic in zoo.cfg
+  sed -i 's/dynamicConfigFile=.*/dynamicConfigFile=\/data\/zoo\.cfg\.dynamic/g' $ZOOCFGDIR/zoo.cfg
```
spiegela: Let's add a comment here to make the purpose of this replacement clear.  From what I can tell:

```
# This node was found in the zookeeper ensemble on disk, setting dynamic config to the base dynamic config file.
```

Do I understand the purpose of this line correctly?

docker/bin/zookeeperStart.sh
```diff
@@ -99,9 +98,35 @@ if [[ "$WRITE_CONFIGURATION" == true ]]; then
...
-
   fi
 fi
 
+ZOOCFGDIR=/data/conf
+export ZOOCFGDIR
+
+if [ ! -d "$ZOOCFGDIR" ]
+then
+echo "Creating writable conf directory."
```
spiegela: I feel like we should make a comment here to make the reasoning for this more explicit:

```
# Copying configmap generated contents to writable directory, to support Zookeeper dynamic reconfiguraiton
```

Or some such.

docker/bin/zookeeperStart.sh
```diff
@@ -99,9 +98,35 @@ if [[ "$WRITE_CONFIGURATION" == true ]]; then
...
+fi
+
+if [[ "$WRITE_CONFIGURATION" == false &&  "$REGISTER_NODE" == false ]]; then
+  # We get here only on server restart...
+  echo Printing bootstrap dynamic config file
+  cat $DYNCONFIG
+  echo Printing Static Config
+  STATIC_CONFIG=`cat $ZOOCFGDIR/zoo.cfg`
+  echo Writing configuration gleaned from zookeeper ensemble
+  echo "$CONFIG" | grep -v "^version="> $DYNCONFIG
```
fpj: One clarification question: `$CONFIG` is an input to the script, what's it expected to be, the last dynamic config file?

docker/bin/zookeeperStart.sh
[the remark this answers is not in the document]
pbelgundi: Sure. Will add.

docker/bin/zookeeperStart.sh
pbelgundi: Yup, It sets dynamic configuration file to the current cluster configuration, fetched from zk-client service.

docker/bin/zookeeperStart.sh
pbelgundi: Sure, will add more comments.

docker/bin/zookeeperReady.sh
```diff
@@ -24,48 +24,46 @@ OK=$(echo ruok | nc 127.0.0.1 $CLIENT_PORT)
...
-    exit 1
+    elif [[ "$ROLE" == "observer" ]]; then
 
-  else
-    echo "Something has gone wrong. Unable to determinal zookeeper role."
-    exit 1
+      echo "Zookeeper service is ready to be upgraded from observer to participant."
+      ROLE=participant
+      ZKCONFIG=$(zkConfig)
+      java -Dlog4j.configuration=file:"$LOG4J_CONF" -jar /root/zu.jar remove $ZKURL $MYID
```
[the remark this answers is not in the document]
pbelgundi: On first start, all zk servers, other than the first one (zk0) start as "observers" and this code moves them from "observer" to "participant". "zk0" on the other hand starts as a "participant" on first start. 
On restart, every server starts as "observer". 
This I guess was done for reasons mentioned in the "warning" box here: https://zookeeper.apache.org/doc/r3.5.3-beta/zookeeperReconfig.html#sc_reconfig_modifying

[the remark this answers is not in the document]
pbelgundi: As regards the 1 sec delay, not sure what would be a good value to have there? The probe itself fires every 10 secs....
Also when the remove command returns, if  we can guarantee that the node is removed then the delay may be completely unnecessary...

docker/bin/zookeeperStart.sh
pbelgundi: The value in $CONFIG is generated by querying zk-service for the current dynamic configuration.
So it will always reflect the current dynamic configuration.

docker/bin/zookeeperReady.sh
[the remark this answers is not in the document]
fpj: The reconfiguration requests are totally ordered, so you are right that the sleep shouldn't really matter. I'm wondering if we need the removal at all, can we directly reconfigure it to be a participant?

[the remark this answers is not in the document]
fpj: I created issue #70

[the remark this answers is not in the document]
pbelgundi: Not sure about this.

pbelgundi:
> In addition to the comments, I also don't understand why `Gopkg.lock` is being added in this PR.

This is a generated file and looks like was committed about 2 months back by mistake.
I tried removing the file from my local code base and running build, test and make-build and things work fine without it as it gets regenerated.
I have removed it from git in the last commit.

PR #73, opened 2019-07-04
Issue 72: Reduce Pod recovery time

Tristan1900:
### Change log description
Reduce the timeout waiting for the [`Future`](https://github.com/pravega/zookeeper-operator/blob/88748261ed286f75b043740217ed2e03942f7eac/docker/zu/src/main/java/io/pravega/zookeeper/Zookeeper.kt#L62) to get a result from zk-client. When timeout fires the zk pod will crash and restart to try to connect again. In our experiments, this method can reduce the waiting time. 

### Purpose of the change
Fix #72 

### How to verify it
Create a zk cluster of size 5 in PKS, delete 2 of them forcefully and see how long do these two pods come back up. 

Signed-off-by: wenqimou <452787782@qq.com>

PR #93, opened 2019-10-21
Issue 92: Zookeeper pods fail to restart when quorum not available

### LINKED ISSUE #92: Zookeeper pods fail to come back up when > quorum number of pods are deleted.
Steps to Reproduce:
1. Create a 3 node zookeeper cluster.
2. Once cluster is up and running delete 2 or more pods from the cluster using `kubectl delete`.
3. The new restarted pods are not able to join back the zk cluster becuase of lack of quorum and logs show the following error message:
"
Error performing zookeeper reconfiguration:
org.apache.zookeeper.KeeperException$NewConfigNoQuorum: KeeperErrorCode = NewConfigNoQuorum
at org.apache.zookeeper.KeeperException.create(KeeperException.java:110)
at org.apache.zookeeper.KeeperException.create(KeeperException.java:54)
at org.apache.zookeeper.ZooKeeper.internalReconfig(ZooKeeper.java:3081)
at org.apache.zookeeper.admin.ZooKeeperAdmin.reconfigure(ZooKeeperAdmin.java:174)
at io.pravega.zookeeper.MainKt.reconfigure(Main.kt:99)
at io.pravega.zookeeper.MainKt.runAdd(Main.kt:122)
at io.pravega.zookeeper.MainKt.main(Main.kt:29)
"

pbelgundi:

Change log description
Changed zookeeper startup, shutdown and readiness scripts to deal with server restarts without invoking "reconfig".

Purpose of the change
Fixes #92

What the code does
Changes the way we handle zookeeper server restarts
TDB: Changes to Zookeeper-operator for handling permanent deletion of servers from zookeeper ensemble when we choose to scale down the zk cluster. For addition of new clusters to the ensemble (on scale up), that is already covered by the existing zookeeperStart.sh script.

How to verify it
All tests should pass.
Manually tested this on a 3 node zookeeper cluster.
Tried deleting all 3 nodes, one at a time and they came back up successfully.
Tried deleting 2 nodes at a time and they came back up successfully.
Tried deleting all 3 nodes in the cluster and the cluster was restored back in <5 mins.

### DIFF
--- a/docker/bin/zookeeperReady.sh
+++ b/docker/bin/zookeeperReady.sh
@@ -54,15 +54,20 @@ if [[ "$OK" == "imok" ]]; then
           ONDISK_CONFIG=true
           DYN_CFG_FILE_LINE=`cat $STATIC_CONFIG|grep "dynamicConfigFile\="`
           DYN_CFG_FILE=${DYN_CFG_FILE_LINE##dynamicConfigFile=}
-          SERVER_LINE=`cat $DYN_CFG_FILE | grep "server.${MYID}="`
-          if [[ $SERVER_LINE == *"participant"* ]]; then
+          SERVER_FOUND=`cat $DYN_CFG_FILE | grep "server.${MYID}=" | wc -l`
+          if [[ "$SERVER_FOUND" == "0" ]]; then
+            echo "Server not found in ensemble. Exiting ..."
+            exit 0
+          fi
+          SERVER=`cat $DYN_CFG_FILE | grep "server.${MYID}="`
+          if [[ $SERVER == *"participant"* ]]; then
               ROLE=participant
           else
               ROLE=observer
           fi
       fi
     fi
-    
+
     if [[ "$ROLE" == "participant" ]]; then
       echo "Zookeeper service is available and an active participant"
     exit 0
--- a/docker/bin/zookeeperReady.sh
+++ b/docker/bin/zookeeperReady.sh
@@ -24,7 +24,6 @@ OK=$(echo ruok | nc 127.0.0.1 $CLIENT_PORT)
 
 # Check to see if zookeeper service answers
 if [[ "$OK" == "imok" ]]; then
-
   set +e
   nslookup $DOMAIN
   if [[ $? -eq 1 ]]; then
@@ -36,7 +35,7 @@ if [[ "$OK" == "imok" ]]; then
     # An ensemble exists, check to see if this node is already a member.
     # Check to see if zookeeper service for this node is a participant
     set +e
-    # Extract resource name and this members ordinal value from pod hostname
+    # Extract resource name and this members' ordinal value from pod hostname
     HOST=`hostname -s`
     if [[ $HOST =~ (.*)-([0-9]+)$ ]]; then
         NAME=${BASH_REMATCH[1]}
@@ -50,19 +49,19 @@ if [[ "$OK" == "imok" ]]; then
     if [ -f $MYID_FILE ]; then
       EXISTING_ID="`cat $DATA_DIR/myid`"
       if [[ "$EXISTING_ID" == "$MYID" && -f $STATIC_CONFIG ]]; then
-        # If Id is correct and configuration is present under `/data/conf`
+      #If Id is correct and configuration is present under `/data/conf`
           ONDISK_CONFIG=true
           DYN_CFG_FILE_LINE=`cat $STATIC_CONFIG|grep "dynamicConfigFile\="`
           DYN_CFG_FILE=${DYN_CFG_FILE_LINE##dynamicConfigFile=}
           SERVER_FOUND=`cat $DYN_CFG_FILE | grep "server.${MYID}=" | wc -l`
           if [[ "$SERVER_FOUND" == "0" ]]; then
             echo "Server not found in ensemble. Exiting ..."
-            exit 0
+            exit 1
           fi
           SERVER=`cat $DYN_CFG_FILE | grep "server.${MYID}="`
-          if [[ $SERVER == *"participant"* ]]; then
+          if [[ ""$SERVER" == *"participant"* ]]; then
               ROLE=participant
-          else
+          elif [[ "$SERVER" == *"observer"* ]]; then
               ROLE=observer
           fi
       fi
--- a/docker/bin/zookeeperReady.sh
+++ b/docker/bin/zookeeperReady.sh
@@ -18,6 +18,7 @@ HOST=`hostname -s`
 DATA_DIR=/data
 MYID_FILE=$DATA_DIR/myid
 LOG4J_CONF=/conf/log4j-quiet.properties
+STATIC_CONFIG=/data/conf/zoo.cfg
 
 OK=$(echo ruok | nc 127.0.0.1 $CLIENT_PORT)
 
@@ -27,40 +28,57 @@ if [[ "$OK" == "imok" ]]; then
   set +e
   nslookup $DOMAIN
   if [[ $? -eq 1 ]]; then
-
     set -e
     echo "There is no active ensemble, skipping readiness probe..."
     exit 0
-
   else
     set -e
     # An ensemble exists, check to see if this node is already a member.
     # Check to see if zookeeper service for this node is a participant
     set +e
-    ZKURL=$(zkConnectionString)
-    set -e
-    MYID=`cat $MYID_FILE`
-    ROLE=`java -Dlog4j.configuration=file:"$LOG4J_CONF" -jar /root/zu.jar get-role $ZKURL $MYID`
-
+    # Extract resource name and this members ordinal value from pod hostname
+    HOST=`hostname -s`
+    if [[ $HOST =~ (.*)-([0-9]+)$ ]]; then
+        NAME=${BASH_REMATCH[1]}
+        ORD=${BASH_REMATCH[2]}
+    else
+        echo Failed to parse name and ordinal of Pod
+        exit 1
+    fi
+    MYID=$((ORD+1))
+    ONDISK_CONFIG=false
+    if [ -f $MYID_FILE ]; then
+      EXISTING_ID="`cat $DATA_DIR/myid`"
+      if [[ "$EXISTING_ID" == "$MYID" && -f $STATIC_CONFIG ]]; then
+        # If Id is correct and configuration is present under `/data/conf`
+          ONDISK_CONFIG=true
+          DYN_CFG_FILE_LINE=`cat $STATIC_CONFIG|grep "dynamicConfigFile\="`
+          DYN_CFG_FILE=${DYN_CFG_FILE_LINE##dynamicConfigFile=}
+          SERVER_LINE=`cat $DYN_CFG_FILE | grep "server.${MYID}="`
+          if [[ $SERVER_LINE == *"participant"* ]]; then
+              ROLE=participant
+          else
+              ROLE=observer
+          fi
+      fi
+    fi
+    
     if [[ "$ROLE" == "participant" ]]; then
       echo "Zookeeper service is available and an active participant"
-      exit 0
-
+    exit 0
     elif [[ "$ROLE" == "observer" ]]; then
-
       echo "Zookeeper service is ready to be upgraded from observer to participant."
       ROLE=participant
+      ZKURL=$(zkConnectionString)
       ZKCONFIG=$(zkConfig)
       java -Dlog4j.configuration=file:"$LOG4J_CONF" -jar /root/zu.jar remove $ZKURL $MYID
       sleep 1
       java -Dlog4j.configuration=file:"$LOG4J_CONF" -jar /root/zu.jar add $ZKURL $MYID $ZKCONFIG
       exit 0
-
     else
       echo "Something has gone wrong. Unable to determine zookeeper role."
       exit 1
     fi
-
   fi
 
 else
--- a/docker/bin/zookeeperReady.sh
+++ b/docker/bin/zookeeperReady.sh
@@ -59,7 +59,7 @@ if [[ "$OK" == "imok" ]]; then
             exit 1
           fi
           SERVER=`cat $DYN_CFG_FILE | grep "server.${MYID}="`
-          if [[ ""$SERVER" == *"participant"* ]]; then
+          if [[ "$SERVER" == *"participant"* ]]; then
               ROLE=participant
           elif [[ "$SERVER" == *"observer"* ]]; then
               ROLE=observer
@@ -69,7 +69,7 @@ if [[ "$OK" == "imok" ]]; then
 
     if [[ "$ROLE" == "participant" ]]; then
       echo "Zookeeper service is available and an active participant"
-    exit 0
+      exit 0
     elif [[ "$ROLE" == "observer" ]]; then
       echo "Zookeeper service is ready to be upgraded from observer to participant."
       ROLE=participant
--- a/docker/bin/zookeeperStart.sh
+++ b/docker/bin/zookeeperStart.sh
@@ -68,7 +68,6 @@ if [[ "$ACTIVE_ENSEMBLE" == false ]]; then
   # This is the first node being added to the cluster
   REGISTER_NODE=false
 else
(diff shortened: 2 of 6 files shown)


pbelgundi:
For dealing with removal of servers from zk ensemble on a scale down, tried the following approach:
Added a "CLUSTER_SIZE" field to the zookeeper config map, which should be equal in value to `spec.Replicas`. When the cluster is being scaled down by reducing the number of replicas, the "CLUSTER_SIZE" field in config map should be updated before the sts for zookeeper is updated with new number of replicas.
In `zookeeperTeardown.sh` script added this line:
``
if [[ "$CLUSTER_SIZE" -lt "$MYID" ]]; then
  # If ClusterSize < MyId, this server is being permanantly removed....
 # so remove it from the ensemble
  echo $?
fi
```

However this does not seem to work consistenly, due to updation issue with Config Maps in kubernetes:
https://kubernetes.io/docs/tasks/configure-pod-container/configure-pod-configmap/#mounted-configmaps-are-updated-automatically

**Comment by pbelgundi** [comment]
As per k8s documentation, there is no upper time limit or deterministic duration within which a config map can always be expected to have the updated value. 
```
As a result, the total delay from the moment when the ConfigMap is updated to the moment when new keys are projected to the pod can be as long as kubelet sync period (1 minute by default) + ttl of ConfigMaps cache (1 minute by default) in kubelet.
```
This makes it risky to use value of `CLUSTER_SIZE` in config map, since using an old value would lead to the extra servers never getting deleted.

**Comment by pbelgundi** [comment]
Tested changes by deleting nodes in all combinations and scaling the # of replicas up and down for the zookeeper cluster.
After every change checked the cluster state and ensemble config and it did get updated as expected.

File: docker/bin/zookeeperReady.sh [hunk 1]
```diff
@@ -27,40 +28,62 @@ if [[ "$OK" == "imok" ]]; then
...
+      EXISTING_ID="`cat $DATA_DIR/myid`"
+      if [[ "$EXISTING_ID" == "$MYID" && -f $STATIC_CONFIG ]]; then
+        # If Id is correct and configuration is present under `/data/conf`
+          ONDISK_CONFIG=true
+          DYN_CFG_FILE_LINE=`cat $STATIC_CONFIG|grep "dynamicConfigFile\="`
+          DYN_CFG_FILE=${DYN_CFG_FILE_LINE##dynamicConfigFile=}
+          SERVER_FOUND=`cat $DYN_CFG_FILE | grep "server.${MYID}=" | wc -l`
+          if [[ "$SERVER_FOUND" == "0" ]]; then
+            echo "Server not found in ensemble. Exiting ..."
+            exit 0
```
spiegela: @pbelgundi as we discussed in slack, this should be an exit code of `1` indicating that this node is _not ready_ since its not in the ensemble.

pkg/controller/zookeepercluster/zookeepercluster_controller.go
```diff
@@ -185,20 +188,71 @@ func (r *ReconcileZookeeperCluster) reconcileStatefulSet(instance *zookeeperv1be
...
+		Name:      cm.Name,
+		Namespace: cm.Namespace,
+	}, foundCm)
+	if err != nil {
+		r.log.Error(err, "Error getting config map.")
+		return false
+	} else {
+		// found config map, now check numer of replicas in configMap
+		envStr := foundCm.Data["env.sh"]
+		r.log.Info("---ENV.SH---", "String", envStr)
```
spiegela: Looks like this is here for debugging?  Should it be removed before merge?

pkg/controller/zookeepercluster/zookeepercluster_controller.go
```diff
@@ -185,20 +188,71 @@ func (r *ReconcileZookeeperCluster) reconcileStatefulSet(instance *zookeeperv1be
...
+	foundCm := &corev1.ConfigMap{}
+	err := r.client.Get(context.TODO(), types.NamespacedName{
+		Name:      cm.Name,
+		Namespace: cm.Namespace,
+	}, foundCm)
+	if err != nil {
+		r.log.Error(err, "Error getting config map.")
+		return false
+	} else {
+		// found config map, now check numer of replicas in configMap
```
fpj: Typo: `number`

pkg/controller/zookeepercluster/zookeepercluster_controller.go
```diff
@@ -185,20 +188,71 @@ func (r *ReconcileZookeeperCluster) reconcileStatefulSet(instance *zookeeperv1be
...
+		splitSlice := strings.Split(envStr, "CLUSTER_SIZE=")
+		if len(splitSlice) < 2 {
+			r.log.Error(err, "Error: Could not find cluster size in configmap.")
+			return false
+		}
+		r.log.Info("---SIZE VALUE---", "SplitString", splitSlice[1])
+		cs := strings.TrimSpace(splitSlice[1])
+		r.log.Info("---CS---", "CSKEY", cs)
+		clusterSize, _ := strconv.Atoi(cs)
+		r.log.Info("--CLUSTER_SIZE---", "CLUSTER_SIZE:", clusterSize)
```
fpj: You made this comment in the conversation tab:

> When spec.Replicas is updated, the "CLUSTER_SIZE" field in config map would also be updated with the same value.

If the two variables hold the same value and mean pretty much the same thing, then why do we need both?

pkg/controller/zookeepercluster/zookeepercluster_controller.go
```diff
@@ -185,20 +188,71 @@ func (r *ReconcileZookeeperCluster) reconcileStatefulSet(instance *zookeeperv1be
...
-		err = r.client.Update(context.TODO(), foundSts)
-		if err != nil {
-			return err
+		foundSTSSize := *foundSts.Spec.Replicas
+		newSTSSize := *sts.Spec.Replicas
+		if newSTSSize < foundSTSSize {
+			// We're dealing with STS scale down
+			if !r.isConfigMapInSync(instance) {
+				r.log.Info("Skipping StatefulSet reconcile as ConfigMap not updated yet.")
+				return nil
```
fpj: In the case that it is not reconciled, we exit, and I assume we retry this function eventually? what's the mechanism for retrying?

docker/bin/zookeeperReady.sh
```diff
@@ -27,40 +28,62 @@ if [[ "$OK" == "imok" ]]; then
...
   else
     set -e
     # An ensemble exists, check to see if this node is already a member.
     # Check to see if zookeeper service for this node is a participant
     set +e
-    ZKURL=$(zkConnectionString)
-    set -e
-    MYID=`cat $MYID_FILE`
-    ROLE=`java -Dlog4j.configuration=file:"$LOG4J_CONF" -jar /root/zu.jar get-role $ZKURL $MYID`
+    # Extract resource name and this members ordinal value from pod hostname
```
fpj: Typo: `member ordinal value`

docker/bin/zookeeperStart.sh
```diff
@@ -31,101 +32,86 @@ fi
...
+    WRITE_CONFIGURATION=false
+else
     WRITE_CONFIGURATION=true
+fi
 
+if [[ "$ACTIVE_ENSEMBLE" == false ]]; then
+  # This is the first node being added to the cluster
+  REGISTER_NODE=false
+else
+  # ACTIVE_ENSEMBLE == true
```
fpj: Is this correctly commented out? If so, remove the line.

pkg/controller/zookeepercluster/zookeepercluster_controller.go
```diff
@@ -185,20 +188,71 @@ func (r *ReconcileZookeeperCluster) reconcileStatefulSet(instance *zookeeperv1be
...
-		r.log.Info("Updating StatefulSet",
-			"StatefulSet.Namespace", foundSts.Namespace,
-			"StatefulSet.Name", foundSts.Name)
-		zk.SyncStatefulSet(foundSts, sts)
-		err = r.client.Update(context.TODO(), foundSts)
-		if err != nil {
-			return err
+		foundSTSSize := *foundSts.Spec.Replicas
+		newSTSSize := *sts.Spec.Replicas
+		if newSTSSize < foundSTSSize {
```
fpj: What about scale up events? Don't we need to handle them here?

docker/zu/src/main/java/io/pravega/zookeeper/Zookeeper.kt
```diff
@@ -17,21 +17,7 @@ import org.apache.zookeeper.admin.ZooKeeperAdmin
...
-    System.err.println("Connecting to Zookeeper $zkUrl")
-    val connectionWatcher = ConnectionWatcher()
-    val zk = ZooKeeper(zkUrl, 3000, connectionWatcher)
-
-    connectionWatcher.waitUntilConnected()
-
-    return zk
-}
- */
+const val ZK_CONNECTION_TIMEOUT_MINS: Long  = 3
```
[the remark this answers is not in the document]
pbelgundi: This change is not needed for this fix to work and restarts do work as expected even if this interval is left unchanged.
Yes, I brought it down to 3 mins from 10 based on #93

pkg/controller/zookeepercluster/zookeepercluster_controller.go
```diff
@@ -185,20 +188,71 @@ func (r *ReconcileZookeeperCluster) reconcileStatefulSet(instance *zookeeperv1be
...
+			// We're dealing with STS scale down
+			if !r.isConfigMapInSync(instance) {
+				r.log.Info("Skipping StatefulSet reconcile as ConfigMap not updated yet.")
+				return nil
+			}
+			//https://kubernetes.io/docs/tasks/configure-pod-container/configure-pod-configmap/#mounted-configmaps-are-updated-automatically
```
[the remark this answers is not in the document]
pbelgundi: Will add text to explain the reconcile skip for the statefulset, but the link is necessary for understanding how long we need to wait.

docker/bin/zookeeperStart.sh
```diff
@@ -31,101 +32,86 @@ fi
...
 WRITE_CONFIGURATION=true
 REGISTER_NODE=true
+ONDISK_CONFIG=false
 
 # Check validity of on-disk configuration
 if [ -f $MYID_FILE ]; then
   EXISTING_ID="`cat $DATA_DIR/myid`"
-  if [[ "$EXISTING_ID" == "$MYID" || -f $DYNCONFIG ]]; then
-      WRITE_CONFIGURATION=false
+  if [[ "$EXISTING_ID" == "$MYID" && -f $STATIC_CONFIG ]]; then
```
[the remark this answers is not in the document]
pbelgundi: Variable `$DYNCONFIG` points to the first DynamicConfigFile "/data/zoo.cfg.dynamic" and its presence does not guarantee that dynamic configuration is present under "/data/conf/" directory Hence changed this check from existence check for "/data/conf/zoo.cfg" instead of "/data/zoo.cfg.dynamic".

pkg/controller/zookeepercluster/zookeepercluster_controller.go
pbelgundi: Because `spec.Replicas` is not visible inside Container scripts and we need that value inside zookeeperTeardown.sh to understand if the pod being terminated is  because of a scale down or a temporary restart.

pkg/controller/zookeepercluster/zookeepercluster_controller.go
pbelgundi: Yes. We will retry it eventually.

docker/bin/zookeeperReady.sh
pbelgundi: This was an existing comment, will change it nevertheless....

pbelgundi: I think it should be the `members' ordinal value`

docker/bin/zookeeperStart.sh
pbelgundi: Ok. Done.

pkg/controller/zookeepercluster/zookeepercluster_controller.go
pbelgundi: No Scale up does not need any special handling.

docker/bin/zookeeperReady.sh
pbelgundi: Done. Changed to exit 1 and tested.

pkg/controller/zookeepercluster/zookeepercluster_controller.go
pbelgundi: Removed debug statements.

docker/bin/zookeeperReady.sh
```diff
@@ -27,40 +28,62 @@ if [[ "$OK" == "imok" ]]; then
...
+    else
+        echo Failed to parse name and ordinal of Pod
+        exit 1
+    fi
+    MYID=$((ORD+1))
+    ONDISK_CONFIG=false
+    if [ -f $MYID_FILE ]; then
+      EXISTING_ID="`cat $DATA_DIR/myid`"
+      if [[ "$EXISTING_ID" == "$MYID" && -f $STATIC_CONFIG ]]; then
+        # If Id is correct and configuration is present under `/data/conf`
```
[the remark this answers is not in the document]
pbelgundi: Fixed in the last commit.

docker/zu/src/main/java/io/pravega/zookeeper/Zookeeper.kt
```diff
@@ -62,4 +48,3 @@ class ConnectionWatcher : Watcher {
         connected.get(ZK_CONNECTION_TIMEOUT_MINS, TimeUnit.MINUTES)
     }
 }
-
```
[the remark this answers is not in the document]
pbelgundi: Ok.

docker/bin/zookeeperReady.sh
pbelgundi: Updated

pbelgundi:
Invalidates PR #73

docker/bin/zookeeperStart.sh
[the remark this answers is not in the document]
fpj: What's it that you are trying to validate here? A static config might exist independent of whether the server has ever joined the ensemble. I thought you're trying to validate that the server has participated in the ensemble previously.

pkg/controller/zookeepercluster/zookeepercluster_controller.go
```diff
@@ -185,20 +188,71 @@ func (r *ReconcileZookeeperCluster) reconcileStatefulSet(instance *zookeeperv1be
...
-			"StatefulSet.Namespace", foundSts.Namespace,
-			"StatefulSet.Name", foundSts.Name)
-		zk.SyncStatefulSet(foundSts, sts)
-		err = r.client.Update(context.TODO(), foundSts)
-		if err != nil {
-			return err
+		foundSTSSize := *foundSts.Spec.Replicas
+		newSTSSize := *sts.Spec.Replicas
+		if newSTSSize < foundSTSSize {
+			// We're dealing with STS scale down
```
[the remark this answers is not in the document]
pbelgundi: Fixed.

docker/bin/zookeeperStart.sh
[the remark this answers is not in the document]
pbelgundi: I'm just trying to validate if this is the first time we're bringing up this server.
In case of a server starting for the very first time, "/data/conf/zoo.cfg" (STATIC_CONFIG) will not exist, as it is copied later in the script to that location. See ln #109 in zookeeperStart.sh.
For first time server startup, we need to explicitly add the server to ensemble by setting REGISTER_NODE=true or incase of zk-0 by adding it directly as 'participant' to the ensemble.
On restarts server is not removed from ensemble, so it should exist in ensemble.

EronWright:
Note to reviewers, this PR caused a critical regression #103.  Please be wary of a) global operator state such as was introduced here (`skipSTSReconcile`), and b) assertions about the timing of the reconcile loop (which may be triggered by watch events, timers, retries, etc).

PR #96, opened 2019-10-31
Issue 94: Fix for zk scale up problem

### LINKED ISSUE #94: Zookeeper instances fail to start on second scale up
## Problem description
Had a pravega installation with 7 zookeeper . 
scaled down zk from 7 to 5 and then 3 .
IO was running fine .No pod restarts were observed.
Later on scaled up from 3 instances of zookeeper to 5 instances .
observed zookeeper pods restarts 
Error observed:

2019-10-17 10:05:58,998 [myid:6] - ERROR [main:QuorumPeerMain@101] - Unexpected exception, exiting abnormally
java.lang.RuntimeException: My id 6 not in the peer list
        at org.apache.zookeeper.server.quorum.QuorumPeer.start(QuorumPeer.java:903)
        at org.apache.zookeeper.server.quorum.QuorumPeerMain.runFromConfig(QuorumPeerMain.java:205)
        at org.apache.zookeeper.server.quorum.QuorumPeerMain.initializeAndRun(QuorumPeerMain.java:123)
        at org.apache.zookeeper.server.quorum.QuorumPeerMain.main(QuorumPeerMain.java:82)

pbelgundi:

## Change log description
This issue occurs because on scale down PVCs attached to the scaled down servers are not deleted.
The fix is about deleting orphaned PVCs once a cluster has been scaled down.
For this fix to take effect `VolumeReclaimPolicy` in zookeeperSpec needs to be set to `Delete`

## Purpose of the change
Fixes #94 

## How to verify it
Verified by scaling the cluster up and down 5 times after this fix and there were no issues with cluster scale and pod restarts. 
Also PVCs attached to old pods were deleted after the scale down.


pkg/apis/zookeeper/v1beta1/zookeepercluster_types.go
```diff
@@ -374,7 +374,8 @@ type Persistence struct {
 func (p *Persistence) withDefaults() (changed bool) {
 	if !p.VolumeReclaimPolicy.isValid() {
 		changed = true
-		p.VolumeReclaimPolicy = VolumeReclaimPolicyRetain
+		// Default volume reclaim policy should be delete
+		p.VolumeReclaimPolicy = VolumeReclaimPolicyDelete
```
spiegela: Why should the default here be to delete?  This is the opposite behavior of a `StatefulSet`, and is not the expected behavior in ECS Flex (which also uses the ZK operator).

pkg/utils/finalizer_utils.go
```diff
@@ -32,3 +37,17 @@ func RemoveString(slice []string, str string) (result []string) {
 	}
 	return result
 }
+
+func IsPvcOrphan(zkPvcName string, replicas int32) bool {
```
spiegela: The acronym should have common capitalization here: `IsPVCOrphan`.

pkg/apis/zookeeper/v1beta1/zookeepercluster_types.go
pbelgundi: @spiegela If the `VolumeReclaimPolicy` is `Retain`, PVCs don't get deleted after the zk cluster is completely deleted (see code in `reconcileFinalizers`), which is not good because if the PVCs keep lying around they cause a problem when cluster is recreated with the same names as these get reattached to the new servers.
Currently to get around this problem, we explicitly set the persistence value in the manifest:
```
apiVersion: "zookeeper.pravega.io/v1beta1"
kind: "ZookeeperCluster"
metadata:
  name: "example"
spec:
  image:
    repository: "pbelgundi/zookeeper"
    tag: "0.2.4"
    pullPolicy: Always
  size: 3
  persistence:
    reclaimPolicy: Delete
```

I don't see why we would want to retain the volumes after deleting the cluster?

pkg/utils/finalizer_utils.go
pbelgundi: Ok. Sure.

pkg/apis/zookeeper/v1beta1/zookeepercluster_types.go
pbelgundi: 1. The variable `VolumeReclaimPolicy` does not impact the zookeeper sts in any way. It is simply used to understand the users preference to delete/retain the PVCs after deleting the cluster.
2. k8s does not `auto delete` volumes attached to StatefulSets, when they are deleted because they `may` be needed later. So deleting them is left to user discretion based on the use case. In this case having the PVCs lying around does not serve any purpose and in fact causes issues like the one I mentioned earlier.
3. In case a user wants to resurrect a deleted zookeeper cluster, that is still doable by explicitly setting `persistence: retain` in the manifest. But I don't think most production users would want that.

pbelgundi: Changed default back to "Retain". With this, zk manifest used by pravega needs to explicitly mention "persistence: Delete" for this fix to take effect.

pkg/utils/finalizer_utils.go
pbelgundi: Done

pkg/controller/zookeepercluster/zookeepercluster_controller.go
```diff
@@ -559,17 +592,30 @@ func (r *ReconcileZookeeperCluster) cleanUpZookeeperPVC(instance *zookeeperv1bet
...
-		pvcDelete := &corev1.PersistentVolumeClaim{
-			ObjectMeta: metav1.ObjectMeta{
-				Name:      pvcItem.Name,
-				Namespace: pvcItem.Namespace,
-			},
-		}
-		err = r.client.Delete(context.TODO(), pvcDelete)
-		if err != nil {
-			return err
+		if all {
```
spiegela: A conditional switch statement can be used to clean up this nested if / else block.

fpj: Could we instead have separate functions for clean all and clean orphan? I find the separation of functionality based on a flag error prone.

pbelgundi: Sure. Will do this.

pbelgundi: Fixed.

PR #110, opened 2019-12-17
Issue 102: Make zookeeper istio friendly

sylvainOL:
as said in https://github.com/istio/istio/issues/19280#issuecomment-560542530, zookeeper doesn't listen on 0.0.0.0 per default and then is not "service mesh friendly".

Adding `quorumListenOnAllIPs` to `true` makes it listen to all ports and not one IP address.

Signed-off-by: Sylvain Desbureaux <sylvain.desbureaux@orange.com>


spiegela:
@pbelgundi Do you care to review from a Pravega perspective?  This change makes sense to me on the ECS side.

pbelgundi:
This change is being handled by PR #113

Issue #104, opened 2019-12-04
Scale Down may not wait long enough before destroying PODs


maddisondavid:
The Scale Down logic currently waits for the ConfigMap to be updated and then 6 reconcile loops before updating the Statefulset.  The reconcile Loop wait I suspect is a way of waiting until the PODs have received the new ConfigMap environment variables.

The problem is that this sync be up to 1 minute (by default)
>As a result, the total delay from the moment when the ConfigMap is updated to the moment when new keys are projected to the pod can be as long as kubelet sync period (1 minute by default) + ttl of ConfigMaps cache (1 minute by default) in kubelet.

https://kubernetes.io/docs/tasks/configure-pod-container/configure-pod-configmap/#mounted-configmaps-are-updated-automatically 

Just waiting 6 reconcile loops does not seem a valid indicator of time as the reconcile loop can be called multiple times depending on what watches have fired.

**Problem Location**
https://github.com/pravega/zookeeper-operator/blob/master/pkg/controller/zookeepercluster/zookeepercluster_controller.go#L207-L210

pbelgundi:
Fixed by PR: https://github.com/pravega/zookeeper-operator/pull/120

PR #111, opened 2019-12-19
Changes for zk operator to be stateless

### LINKED ISSUE #103: Scale Down doesn't work with multiple ZK Clusters
During Scale Down the ConfigMap is updated with a new `CLUSTER_SIZE`, there's a wait (6 reconcile loops) and then the StatefulSet replicas are updated.  This triggers the `zookeeperTeardown.sh` script when the StatefulSet attempts to to remove the POD which in turn allows the dying node to be gracefully removed from the ensemble.

The problem is that the number of reconcile loops is stored in Operator state, so if the operator is managing multiple clusters some of them will trigger prematurely which will result in PODs being destroyed without being gracefully removed.

**Problem Location**
`skipSTSReconcile` should not be stored in controller state (operators should really be stateless)
https://github.com/pravega/zookeeper-operator/blob/cea93eeb758e14e9b9b9499ba9755026d728b72c/pkg/controller/zookeepercluster/zookeepercluster_controller.go#L117

pbelgundi:

## Change log description
Removed `skipSTSReconcile` since it was a global variable that broke the stateless nature of zookeeper operator. Added a sleep of 2 minutes to fix the config map sync problem.
Considered/tried multiple alternatives like 
1. Update an annotation on the STS after configmap is updated so it triggers the volume sync: The problem with this approach is that when K8s restarts pods for the annotation update to take effect, the server when going down (_in teardown.sh_) is removed from ensemble and hence when it comes back up is not able to join the ensemble again.
2. Remove scaled down servers in readiness/liveness probe of existing servers: Update cluster_size in ZK config-map. Do not remove the scaled down servers in the tear-down script but remove them in a readiness probe of some existing zk server. This approach works fine during a scale down, but during a "scale-up", if a new server comes up and is added to ensemble config, but config-map is not updated yet with the new cluster size, then the readiness probe for an existing server deletes this new server assuming it was left over from last scale down.
3. Use a variable to skip certain number of reconcile attempts after configmap is updated with scaled down replica count: Adding skipSTSReconcile variable to ZookeeperClusterStatus and counting on a certain number of skipped reconciles before updating the STS with the new replica count does not guarantee that we will wait for 2 minutes before updating the STS, since reconcile can get invoked for reasons other than the reconcile time expiring. 

Hence finally choose this simple approach.

## Purpose of the change
Fixes #103 #104 

## How to verify it
Verified by scaling the cluster up and down multiple times and there were no issues with cluster scale and pod restarts.

### DIFF
--- a/pkg/controller/zookeepercluster/zookeepercluster_controller.go
+++ b/pkg/controller/zookeepercluster/zookeepercluster_controller.go
@@ -111,10 +111,9 @@ var _ reconcile.Reconciler = &ReconcileZookeeperCluster{}
 type ReconcileZookeeperCluster struct {
 	// This client, initialized using mgr.Client() above, is a split client
 	// that reads objects from the cache and writes to the apiserver
-	client           client.Client
-	scheme           *runtime.Scheme
-	log              logr.Logger
-	skipSTSReconcile int
+	client client.Client
+	scheme *runtime.Scheme
+	log    logr.Logger
 }
 
 type reconcileFun func(cluster *zookeeperv1beta1.ZookeeperCluster) error
@@ -203,15 +202,13 @@ func (r *ReconcileZookeeperCluster) reconcileStatefulSet(instance *zookeeperv1be
 				For details see:
 				https://kubernetes.io/docs/tasks/configure-pod-container/configure-pod-configmap/#mounted-configmaps-are-updated-automatically
 			*/
-			r.skipSTSReconcile++
-			if r.skipSTSReconcile < 6 {
-				r.log.Info("Waiting for Config Map update to sync...Skipping STS Reconcile")
-				return nil
+			for i := 0; i < 4; i++ {
+				time.Sleep(30 * time.Second)
+				r.log.Info("Waiting for Config Map update to sync before STS Reconcile")
 			}
 		}
 		return r.updateStatefulSet(instance, foundSts, sts)
 	}
-	return nil
 }
 
 func (r *ReconcileZookeeperCluster) updateStatefulSet(instance *zookeeperv1beta1.ZookeeperCluster, foundSts *appsv1.StatefulSet, sts *appsv1.StatefulSet) (err error) {
@@ -225,7 +222,6 @@ func (r *ReconcileZookeeperCluster) updateStatefulSet(instance *zookeeperv1beta1
 	}
 	instance.Status.Replicas = foundSts.Status.Replicas
 	instance.Status.ReadyReplicas = foundSts.Status.ReadyReplicas
-	r.skipSTSReconcile = 0
 	return nil
 }
 
--- a/docker/bin/zookeeperStart.sh
+++ b/docker/bin/zookeeperStart.sh
@@ -88,8 +88,12 @@ if [[ "$WRITE_CONFIGURATION" == true ]]; then
     echo $MYID > $MYID_FILE
     echo "server.${MYID}=${ZKCONFIG}" > $DYNCONFIG
   else
+    set -e
+    ZKURL=$(zkConnectionString)
+    CONFIG=`java -Dlog4j.configuration=file:"$LOG4J_CONF" -jar /root/zu.jar get-all $ZKURL`
     echo Writing configuration gleaned from zookeeper ensemble
     echo "$CONFIG" | grep -v "^version="> $DYNCONFIG
+    set +e
   fi
 fi
 
@@ -97,8 +101,10 @@ if [[ "$REGISTER_NODE" == true ]]; then
     ROLE=observer
     ZKURL=$(zkConnectionString)
     ZKCONFIG=$(zkConfig)
+    set -e
     echo Registering node and writing local configuration to disk.
     java -Dlog4j.configuration=file:"$LOG4J_CONF" -jar /root/zu.jar add $ZKURL $MYID  $ZKCONFIG $DYNCONFIG
+    set +e
 fi
 
 ZOOCFGDIR=/data/conf


jkhalack:
Was the solution verified for the case of zk-operator handling multiple zk clusters that are deployed in **more than one namespace** at the same time?

pkg/controller/zookeepercluster/zookeepercluster_controller.go
```diff
@@ -203,15 +202,13 @@ func (r *ReconcileZookeeperCluster) reconcileStatefulSet(instance *zookeeperv1be
 				For details see:
 				https://kubernetes.io/docs/tasks/configure-pod-container/configure-pod-configmap/#mounted-configmaps-are-updated-automatically
 			*/
-			r.skipSTSReconcile++
-			if r.skipSTSReconcile < 6 {
-				r.log.Info("Waiting for Config Map update to sync...Skipping STS Reconcile")
-				return nil
+			for i := 0; i < 4; i++ {
+				time.Sleep(30 * time.Second)
```
fpj: Do we really need to sleep 4x for 30 seconds and print the same log message 4x? It would make sense to do it in the case we are checking whether the config map has been updated.

pkg/controller/zookeepercluster/zookeepercluster_controller.go
```diff
@@ -203,15 +202,13 @@ func (r *ReconcileZookeeperCluster) reconcileStatefulSet(instance *zookeeperv1be
 				For details see:
 				https://kubernetes.io/docs/tasks/configure-pod-container/configure-pod-configmap/#mounted-configmaps-are-updated-automatically
 			*/
-			r.skipSTSReconcile++
-			if r.skipSTSReconcile < 6 {
-				r.log.Info("Waiting for Config Map update to sync...Skipping STS Reconcile")
-				return nil
+			for i := 0; i < 4; i++ {
```
fpj: Why did you choose 2 minutes? Can it be less? In any case, let's not conflate issues. Issue #103 is about removing the global state (through skipSTSReconcile), which this PR does. it is separate question, but not less important, whether we can drop the wait time or even avoid it altogether.

This mechanism for notifying a scale-down is not very elegant, and we should look into replacing it with one that is more reliable and does not induce long wait times.

pkg/controller/zookeepercluster/zookeepercluster_controller.go
pbelgundi: We come to the for loop (at ln #205) only after the config map has been updated with the new value and that check is done by this piece of code a few lines earlier:
"
if !r.isConfigMapInSync(instance) {
"
The for loop is to wait for the volume to be updated as there is no way to check this directly by opening the config map volume on each pod in the STS. 
The reason we wait for 2 minutes is based on this:
https://kubernetes.io/docs/tasks/configure-pod-container/configure-pod-configmap/#mounted-configmaps-are-updated-automatically

pkg/controller/zookeepercluster/zookeepercluster_controller.go
pbelgundi: `skipSTSReconcile` was added so that we could wait for the volume to be updated ....so this is directly related to scale down.
2 mins is the max sync time expected by kubernetes for volume updates based on this:
https://kubernetes.io/docs/tasks/configure-pod-container/configure-pod-configmap/#mounted-configmaps-are-updated-automatically

I agree that this is not a 100% reliable way to handle scale down, but it should work almost always and all the other approaches I have tried did not work due to kubernetes operational dynamics (see PR description). The worst case down side of this fix is that the deleted node would continue to stay in the ensemble and would have to be manually deleted by the admin.

Storing CLUSTER_SIZE in a zookeeper node feels like a little very complicated solution for handling a simple scale down and also introduces additional external dependency. It would require storing one value per cluster handled by the zk-operator.

pkg/controller/zookeepercluster/zookeepercluster_controller.go
spiegela: Hmm.  We should avoid long blocking operations in the reconcile loop, since they may disrupt the operator from from handling changes when running multiple ZK clusters with one operator.

Perhaps we can make a go-client call to the config-map in the container's liveness check rather than relying on teh file-system mount?  This would require adding an addition client to the container image...

Alternatively, if we must really wait 2 minutes, we could store a new field in the ZookeeperCluster Status block, and allow the periodic reconcile loop to replace the sleeps.

pbelgundi: Will check on getting these options to work....

pkg/controller/zookeepercluster/zookeepercluster_controller.go
fpj: > skipSTSReconcile was added so that we could wait for the volume to be updated ....so this is directly related to scale down.

Ok, but let's agree that the main reason for having this PR is to avoid a global variable like `skipSTSReconcile`, which causes problems in the presence of multiple zk clusters. If my understanding is incorrect, then I 

> I agree that this is not a 100% reliable way to handle scale down

Can you enumerate the low-probability cases in which it fails? As part of answering this, let's step and review and what's happening:

1- In the case a zk cluster is being scaled-down, there is a zookeeper reconfig command executed against the zk ensemble. That's the first thing that happens.
2- The second step is to reconfigure the stateful set to have the correct number of replicas (pods).

If 1 and 2 are correct, including the order, then the zk ensemble will in principle work fine after 1, but step 2 might take time to execute, which is fine because it is not safety issue, it is just taking time to execute.

I'm also wondering what happens in the case of multiple consecutive scale down events (e.g., two severs are removed separately and consecutively).

pkg/controller/zookeepercluster/zookeepercluster_controller.go
pbelgundi: As suggested, tried installing `kubectl` on the zk container and using `kubectl describe <configmap>` to get the value of `CLUSTER_SIZE` variable inside the pod, instead of reading from volume. But it did not work because kubectl command failed in the absence of the right service account, roles and role-bindings being present for accessing the resource.

pbelgundi:
The fix was instead implemented by PR: https://github.com/pravega/zookeeper-operator/pull/120

PR #112, opened 2019-12-27
Issue 108: Removing size field

### LINKED ISSUE #108: Remove size field from zookeeper operator
Currently there are 2 fields `size` and `replicas` that serve the same purpose.
Size was deprecated about 10 months ago, but is still mentioned in the documentation and may be used in some places in the code.
We need to have just 1 field for cluster size to avoid confusion and bad user experience.
Charts also need to be updated to use the new `replicas` field.

SrishT:

### Change log description
Removing reference to the field `size` from code and documentation which has been depracated and replaced by the field `replicas`.

### Purpose of the change
Fixes #108 

### How to verify it
The code sanity was tested by deploying a zookeeper cluster and by upscaling and downscaling it. When the field `size` was used instead of `replicas` in the manifest file, it did not throw any error, rather it created the zookeeper cluster with the default value of the field.


pkg/apis/zookeeper/v1beta1/zookeepercluster_types.go
```diff
@@ -45,12 +45,6 @@ type ZookeeperClusterSpec struct {
 	// the zookeeper cluster.
 	Labels map[string]string `json:"labels,omitempty"`
 
-	// Size (DEPRECATED) is the expected size of the zookeeper cluster. This
```
pbelgundi: Removing a field from the ZookeeperCluster object would require regenerating the deep-copy class - `zz_generated.deepcopy.go`. I don't see that being done.

SrishT: done

PR #113, opened 2020-01-01
Issue 102: Add support for ZK to be able to run on Istio enabled cluster

baluchicken:
This PR adds support to enable `quorumListenOnAllIPs`. This config is required for Zookeeper to be able to run on an Istio based environment.
Some modification must be done in the zookeeperStartup script as well, since it uses `nslookup` to determine if the ZkServer is part of an ensemble or not. In case of Istio all the communication goes through the sidecar container which might not ready when the nslookup command issued. So we might need to wait a while to the sidecar to become ready. This script can be removed when Kubernetes 1.17 is more commonly used.

I also modified the Zookeeper docker container slightly, to add `curl`. Also I could not make the official 3.5.5 image work, I am guessing something wrong there, since it also does not work in a non Istio environment.
Fixes: #102

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -14,10 +14,8 @@ COPY zu /zu
 WORKDIR /zu
 RUN ./gradlew assemble
 
-FROM zookeeper:3.5.5
+FROM zookeeper:3.5.4-beta
 COPY bin /usr/local/bin
+RUN apk add curl
 RUN chmod +x /usr/local/bin/*
 COPY --from=0 /zu/build/libs/zu.jar /root/
-
-RUN apt-get -q update && \
-    apt-get install -y dnsutils
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -16,6 +16,8 @@ RUN ./gradlew assemble
 
 FROM zookeeper:3.5.5
 COPY bin /usr/local/bin
-RUN apt update && apt install curl -y
 RUN chmod +x /usr/local/bin/*
 COPY --from=0 /zu/build/libs/zu.jar /root/
+
+RUN apt-get -q update && \
+    apt-get install -y dnsutils curl
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -14,8 +14,8 @@ COPY zu /zu
 WORKDIR /zu
 RUN ./gradlew assemble
 
-FROM zookeeper:3.5.4-beta
+FROM zookeeper:3.5.5
 COPY bin /usr/local/bin
-RUN apk add curl
+RUN apt update && apt install curl -y
 RUN chmod +x /usr/local/bin/*
 COPY --from=0 /zu/build/libs/zu.jar /root/
--- a/docker/bin/zookeeperStart.sh
+++ b/docker/bin/zookeeperStart.sh
@@ -51,6 +51,22 @@ if [ -f $DYNCONFIG ]; then
   ONDISK_DYN_CONFIG=true
 fi
 
+# Check if envoy is up and running
+if [[ "$IS_ISTIO_ENABLED" == true ]]; then
+  COUNT=0
+  MAXCOUNT=${1:-30}
+  HEALTHYSTATUSCODE="200"
+  while true; do
+    COUNT=`expr $COUNT + 1`
+    SC=$(curl -s -o /dev/null -w "%{http_code}" http://localhost:15000/ready)
+    echo "waiting for envoy proxy to come up";
+    sleep 1;
+    if (( "$SC" == "$HEALTHYSTATUSCODE" || "$MAXCOUNT" == "$COUNT" )); then
+      break
+    fi
+  done
+fi
+
 # Determine if there is a ensemble available to join by checking the service domain
 set +e
 nslookup $DOMAIN
--- a/docker/bin/zookeeperStart.sh
+++ b/docker/bin/zookeeperStart.sh
@@ -52,12 +52,12 @@ if [ -f $DYNCONFIG ]; then
 fi
 
 # Check if envoy is up and running
-if [[ "$IS_ISTIO_ENABLED" == true ]]; then
+if [[ -n "$ENVOY_SIDECAR_STATUS" ]]; then
   COUNT=0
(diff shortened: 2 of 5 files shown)


spiegela:
Thanks, @baluchicken.  I was just considering this configuration option while reviewing #110, but after considering it I thought perhaps that `quorumListenOnAllIPs` should just be enabled all the time in K8s configurations (as implemented in #110).

Do you think there's a valid reason to make it configurable?

@fpj @pbelgundi any thoughts on this?

baluchicken:
@spiegela Maybe you are right but, from the Zookeeper documentation I think it shouldn't be enabled all the time since.

```
t affects the connections handling the ZAB protocol and the Fast Leader Election protocol.
```
I assume this means it maybe slows down things.

AceHack:
Any update on this, it's currently blocking us

docker/Dockerfile
```diff
@@ -14,10 +14,8 @@ COPY zu /zu
 WORKDIR /zu
 RUN ./gradlew assemble
 
-FROM zookeeper:3.5.5
+FROM zookeeper:3.5.4-beta
```
jkhalack: Is there a reason to go back to `3.5.4-beta`?

amuraru: @baluchicken I think https://github.com/apache/zookeeper/pull/1254 may address this issue

symat: Hi guy! I created the PR referenced by @amuraru . The issue we are fixing in ZooKeeper with this PR should be present both in 3.5.5 and 3.5.4-beta. So maybe there is an other reason why you want to downgrade to this specific beta version...

baluchicken:
Rebased to master, bumped image to 3.5.5 and removed `isIstioEnabled`, since it can be determined using annotations.

docker/Dockerfile
```diff
@@ -16,8 +16,6 @@ RUN ./gradlew assemble
 
 FROM zookeeper:3.5.5
 COPY bin /usr/local/bin
+RUN apt update && apt install curl -y
 RUN chmod +x /usr/local/bin/*
 COPY --from=0 /zu/build/libs/zu.jar /root/
-
-RUN apt-get -q update && \
-    apt-get install -y dnsutils
```
amuraru: zookeeperStart is using `nslookup` - isn;t that provided by dnsutils?
https://github.com/pravega/zookeeper-operator/blob/20106e883c62d143eaa0c1c26fd7c59f0328dcc8/docker/bin/zookeeperStart.sh#L56

baluchicken: Thanks fixing now.

baluchicken:
@spiegela any update on this? Is there anything needed from my-side? Thanks

baluchicken:
Any update on this? Is there anything needed from my-side? Thanks

pbelgundi:
@baluchicken Regarding:
> In case of Istio all the communication goes through the sidecar container which might not ready when the nslookup command issued. So we might need to wait a while to the sidecar to become ready.

We have already added a wait loop to check nslookup for upto 20 times before bailing out...
https://github.com/pravega/zookeeper-operator/blob/master/docker/bin/zookeeperStart.sh#L63

Your PR branch does not seem to have those changes, so it may help to refresh from master.

pbelgundi:
@baluchicken Could you please explain why you mention this:
> This script can be removed when Kubernetes 1.17 is more commonly used.

pbelgundi:
@baluchicken The current zookeeper docker image we use has the **official 3.5.5** image as the base image, with only additions being startup and health check scripts which are needed by the operator.
So using this image is as good as using the official image.

pbelgundi:
This change is needed perhaps because of this Istio requirement:
https://github.com/istio/istio.io/blob/master/content/en/faq/applications/zookeeper.md

docker/bin/zookeeperStart.sh
```diff
@@ -51,6 +51,22 @@ if [ -f $DYNCONFIG ]; then
   ONDISK_DYN_CONFIG=true
 fi
 
+# Check if envoy is up and running
+if [[ -n "$ENVOY_SIDECAR_STATUS" ]]; then
```
pbelgundi, line 55: The code inside this `If block` shoudl get skipped  if environment variable "ENVOY_SIDECAR_STATUS" is **not** set. 
However, in generators.go we always  set this environment variable, that would cause this block to get executed on non-Istio environments as well.... 
Instead it may be good to inject any required environment variables using a config map, the name of which can be provided in the Zookeeper Spec. 
Similar to : https://github.com/pravega/bookkeeper-operator/blob/master/pkg/apis/bookkeeper/v1alpha1/bookkeepercluster_types.go#L142

baluchicken:
@pbelgundi Thanks for reviewing my PR I will address your findings soon.
I mentioned Kubernetes 1.17 because at the time when I opened this PR the Kubernetes team planned to release something called sidecar container lifecyle. Sadly it cannot make it into 1.17 or 1.18 either it will be available in 1.19(https://banzaicloud.com/blog/k8s-sidecars/).
Using that, the script is no longer required since a way more sophisticated way will be available to check if the sidecar container is ready.

baluchicken:
@pbelgundi I rebased to the master

docker/bin/zookeeperStart.sh
```diff
@@ -51,6 +51,22 @@ if [ -f $DYNCONFIG ]; then
   ONDISK_DYN_CONFIG=true
 fi
 
+# Check if envoy is up and running
```
anishakj, line 54: Can we remove this, since we are  waiting for  40s in nslookup?

baluchicken: For debugging reasons I think it is better to have it separately, and it does not adds more wait time to the startup time. Moreover, it is only affects clusters where Istio is enabled.

Prabhaker24:
@baluchicken I am currently testing your changes I have tested create and delete of the zookeeper cluster with and without Istio enabled. I am still testing for the scaling up and scaling down. I will update here once it's done and then we should be able to merge your pr.

Prabhaker24:
I have tested for scale up and scale down as well with and without Istio both are working.

PR #120, opened 2020-01-28
Issue Changes for handling zk scale down

### LINKED ISSUE #103: Scale Down doesn't work with multiple ZK Clusters

pbelgundi:

## Change log description
Changes to gracefully handle ZK Scale down and remove the stopped servers from the ZK ensemble in the teardown script. The approach taken here is to store the CLUSTER_SIZE on a znode in the same ZK cluster that operator manages and use this value in the teardown script to remove the servers being scaled down.
On startup, a znode with path ` /zookeeper-operator/<zk-cluster-name>` gets created with the value `CLUSTER_SIZE=<number of replicas>` on the zookeeper cluster being managed by the operator. On changing the number of replicas on the zk cluster spec, the same is updated on the znode (_both for scale up as well as scale down_) before being updated on the zookeeper STS. On Scale down, in the teardown script we do a `sync` call on the znode having the  `CLUSTER_SIZE`  and then read this value, to get the most recently updated value for CLUSTER_SIZE.

Removed the global variable `skipSTSReconcile` which messed up multiple cluster management by operator.

## Purpose of the change
Fixes #103 #104 

## How to verify
Tested this by scaling the cluster up and down several times and it worked as expected through all scale ups and downs. The nodes being scaled down were removed from the ensemble by the teardown script.


gaddamas:
Is separate znode created for each cluster zk operator creates/manages?

pbelgundi:
> Is separate znode created for each cluster zk operator creates/manages?

Yes. There is a separate znode for each cluster managed by the operator.

pbelgundi:
> graceful removal of

Currently, in the teardown script, after removing the server from ensemble,  we wait for existing client connections to drain off :
https://github.com/pravega/zookeeper-operator/blob/master/docker/bin/zookeeperTeardown.sh#L33
This is a `preStop ` script and is _blocking_ in nature, which means **delete** on the pod won't be called by K8s till the script completes.
I believe this should take care of the graceful shutdown.

PR #129, opened 2020-02-13
Issue 102: ZK Ensemble fails to bootstrap if headless domain DNS resolution is failing

amuraru:
This patch addresses the issue https://github.com/pravega/zookeeper-operator/issues/132

During bootstrap observer nodes (nodes: 2,...) fail to register if kubernetes headless DNS fails to bootstrap fast enough.
This patch ensures that - in such cases - the dynamic config file is not written, allowing the instance to retry when the failed zk instance is restarted (e.g kubernetes reschedules the pod)


amuraru:
@spiegela @pbelgundi would you please review this? thanks

PR #131, opened 2020-02-28
Issue 124: The additionalPrinterColumn of the Zookeeper Cluster "replicas" prints incorrect value

### LINKED ISSUE #124: Replicas additionalPrinterColumn of the Zookeeper Cluster prints incorrect value
The additional printer column `Replicas` which is part of the Zookeeper Cluster CRD prints the value in `.Status.Replicas` field of the Zookeeper Cluster object which is incorrect. This value should be picked from the object `Spec` and not its `Status`.

SrishT:

### Change log description
The additional printer column `Replicas` which is part of the Zookeeper Cluster CRD prints the value in `.Status.Replicas` field of the Zookeeper Cluster object which is incorrect. This value should be picked from the object Spec and not its Status.

### Purpose of the change
Fixes #124 

### How to verify it
Deployed the Zookeeper Cluster both using charts and manually and confirmed that value printed by the replicas field on doing `kubectl get zk` now prints the correct value.

### DIFF
--- a/charts/zookeeper-operator/Chart.yaml
+++ b/charts/zookeeper-operator/Chart.yaml
@@ -1,6 +1,6 @@
 name: zookeeper-operator
-version: 0.2.5
-appVersion: 0.2.5
+version: latest
+appVersion: latest
 description: zookeeper operator Helm chart for Kubernetes
 keywords:
 - zookeeper
--- a/charts/zookeeper-operator/Chart.yaml
+++ b/charts/zookeeper-operator/Chart.yaml
@@ -1,6 +1,6 @@
 name: zookeeper-operator
-version: 0.2.2
-appVersion: 0.2.2
+version: 0.2.5
+appVersion: 0.2.5
 description: zookeeper operator Helm chart for Kubernetes
 keywords:
 - zookeeper
--- a/charts/zookeeper/Chart.yaml
+++ b/charts/zookeeper/Chart.yaml
@@ -1,6 +1,6 @@
 name: zookeeper
-version: 0.2.2
-appVersion: 0.2.2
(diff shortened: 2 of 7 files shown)


charts/zookeeper-operator/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 name: zookeeper-operator
-version: 0.2.2
-appVersion: 0.2.2
+version: 0.2.5
```
pbelgundi: Please change to `latest` so we don;t have to keep editing this file all the time.

charts/zookeeper/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 name: zookeeper
-version: 0.2.2
-appVersion: 0.2.2
+version: 0.2.5
```
pbelgundi: Same as above

charts/zookeeper/templates/zookeeper.yaml
```diff
@@ -4,3 +4,6 @@ metadata:
   name: {{ template "zookeeper.fullname" . }}
 spec:
   replicas: 3
+  image:
```
pbelgundi: What is this for?

SrishT: This field is needed to use the image mentioned in the values.yaml file

charts/zookeeper-operator/Chart.yaml
SrishT: done

charts/zookeeper/Chart.yaml
SrishT: done

PR #133, opened 2020-03-05
Sync helm chart RBAC with generated files

amuraru:
Helm ClusterRole rules were out of date with the files generated in `deploy`.

### DIFF
--- a/charts/zookeeper-operator/templates/clusterrole.yaml
+++ b/charts/zookeeper-operator/templates/clusterrole.yaml
@@ -13,15 +13,28 @@ rules:
 - apiGroups:
   - ""
   resources:
-  - nodes
+  - pods
+  - services
+  - endpoints
+  - persistentvolumeclaims
+  - events
+  - configmaps
+  - secrets
   verbs:
-  - get
-  - watch
-  - list
-- apiGroups:
-  - admissionregistration.k8s.io
-  resources:
   - "*"
+- apiGroups:
+  - apps
+  resources:
+  - deployments
+  - daemonsets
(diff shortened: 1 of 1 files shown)


amuraru:
@spiegela Aaron - any chance we get this reviewed? Thanks

charts/zookeeper-operator/templates/clusterrole.yaml
```diff
@@ -13,15 +13,28 @@ rules:
 - apiGroups:
   - ""
   resources:
-  - nodes
```
anishakj: why `nodes` is removed from resource list

charts/zookeeper-operator/templates/clusterrole.yaml
```diff
@@ -13,15 +13,28 @@ rules:
...
+  - events
+  - configmaps
+  - secrets
   verbs:
-  - get
-  - watch
-  - list
+  - "*"
 - apiGroups:
-  - admissionregistration.k8s.io
```
anishakj: why `admissionregistration.k8s.io` is removed?

charts/zookeeper-operator/templates/clusterrole.yaml
anishakj: why is `nodes` got removed from resources?

amuraru: it's not needed by the operator - see https://github.com/pravega/zookeeper-operator/blob/master/deploy/all_ns/rbac.yaml

PR #138, opened 2020-03-10
Issue137:  Zookeeper cluster upgrade is failing

### LINKED ISSUE #137: Zookeeper cluster upgrade is failing
Zookeeper cluster upgrade is failing from `0.2.5-rc0` to `0.2.6-rc0` . While trying upgrade pods are reaching `CrashLoopBack` state

anishakj:

## Change log description
Added CLUSTER_SIZE field in configmap

## Purpose of the change
Fixes #137

## How to verify it
 upgrading zookeeper cluster from `0.2.5-rc0` to `0.2.6-rc0` is successful

### DIFF
--- a/pkg/zk/generators.go
+++ b/pkg/zk/generators.go
@@ -18,7 +18,7 @@ import (
 
 	"github.com/pravega/zookeeper-operator/pkg/apis/zookeeper/v1beta1"
 	appsv1 "k8s.io/api/apps/v1"
-	"k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
 	policyv1beta1 "k8s.io/api/policy/v1beta1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime/schema"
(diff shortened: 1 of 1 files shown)


pkg/zk/generators.go
```diff
@@ -226,7 +226,8 @@ func makeZkEnvConfigString(z *v1beta1.ZookeeperCluster) string {
 		"LEADER_PORT=" + strconv.Itoa(int(ports.Leader)) + "\n" +
 		"CLIENT_HOST=" + z.GetClientServiceName() + "\n" +
 		"CLIENT_PORT=" + strconv.Itoa(int(ports.Client)) + "\n" +
-		"CLUSTER_NAME=" + z.GetName() + "\n"
+		"CLUSTER_NAME=" + z.GetName() + "\n" +
+		"CLUSTER_SIZE=" + fmt.Sprint(z.Spec.Replicas) + "\n"
```
pbelgundi: Also update the CLUSTER_SIZE variable in config map on scale up or scale down ( when ever user updates the `zk.spec.replicas` field so that if user triggers an upgrade after doing a scale up/scale down on a zookeeper cluster ( with version 0.2.5) the correct CLUSTER_SIZE is maintained and teardown works as expected.

anishakj: This is taken care by `reConcileConfigmap`. I have tried an upgrade from `0.2.5-rc0` to `0.2.6-rc0` after scaling replicas and it is working fine.

PR #135, opened 2020-03-07
Observers fail to register when zk ensemble service domain is not yet available

amuraru:
### Change log description

When ensemble DOMAIN is not yet available (happens in k8s when svc domain dns resolve is slow to resolve), observer peers (id>=2) fail to register and enter a crash loop.
Previously the DYNCONFIG was populated with peer nodes but `REGISTER_NODE` is false (domain not yet available) hence the `zu add` never called. Subsequently when peer is restarted the registration is never called

### Purpose of the change
 Fixes  #315

### What the code does

DYNCONFIG is not updated, if ensemble is not available for pods starting from ID 2  

### How to verify it
Verified that second zookeeper pod is starting successfully.

PR #143, opened 2020-03-19
Issue134: Support zookeeper cluster setup on minikube

### LINKED ISSUE #134: Support zookeeper cluster setup on minikube
### Description
Current when we try to setup zookeeper cluster using zk-operator on minikube we face issues related to nslookup or other problems.

### Importance
must-have

This JIRA tracks changes necessary to support zookeeper cluster setup on any minikube.

anishakj:
Signed-off-by: anisha.kj anisha.kj@dell.com

## Change log description
In  minikube setup, for zookeeper-1 pod headless service is taking a while to come up and due to that nslookup was failing.  Added polling mechanism to check head less service is active.

## Purpose of the change
Fixes #134 

## How to verify it
Verified multiple times in minikube setup and able to deploy zookeeper  with 3 replicas.

### DIFF
--- a/docker/bin/zookeeperStart.sh
+++ b/docker/bin/zookeeperStart.sh
@@ -56,8 +56,20 @@ set +e
 nslookup $DOMAIN
 if [[ $? -eq 1 ]]; then
   # If an nslookup of the headless service domain fails, then there is no
-  # active ensemble yet
+  # active ensemble yet, but in certain cases nslookup of headless service
+  # takes a while to come up even if there is active ensemble
   ACTIVE_ENSEMBLE=false
+  declare -i count=20
+  while [[ $count -ge 0 && $MYID -ne 1 ]]
+  do
+    sleep 2
+    ((count=count-1))
+    nslookup $DOMAIN
+    if [[ $? -eq 0 ]]; then
+      ACTIVE_ENSEMBLE=true
+      break
+    fi
+  done
 else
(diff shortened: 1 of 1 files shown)


docker/bin/zookeeperStart.sh
```diff
@@ -56,8 +56,20 @@ set +e
...
+  while [[ $count -ge 0 && $MYID -ne 1 ]]
+  do
+    sleep 2
+    ((count=count-1))
+    nslookup $DOMAIN
+    if [[ $? -eq 0 ]]; then
+      ACTIVE_ENSEMBLE=true
+      break
+    fi
+  done
```
amuraru: The retries here are helpful. If it fails though even after 40seconds you may want to review https://github.com/pravega/zookeeper-operator/pull/135 as well. Without that patch the cluster bootstrapping never completes

anishakj:
> LGTM.
> Lets test this extensively to make sure the following functionality works fine post this fix:
> 
> 1. Pod deletions: Delete single, multiple, all zk pods (they should come back up successfully)
> 2. ZK operator and Zk Upgrades.
> 3. ZK Cluster scale up/down.
> 4. Zk Operator managing multiple zk clusters in different namespaces.
> 
> Ideally it would be good to have e2e tests that cover these scenarios.

Tested all scenarios, zk operator upgrade is failing, that is happening even without this change. Raised an issue to track it separately.

PR #151, opened 2020-03-26
Issue150: Helm chart should allow watch namespace to be passed in.

### LINKED ISSUE #150: Helm install should allow WATCH_NAMESPACE to be passed in
Description
Currently the Helm chart for the zookeeper Operator will ONLY watch for resources in the same namespace that the operator is installed:

        env:
        - name: WATCH_NAMESPACE
          valueFrom:
            fieldRef:
              fieldPath: metadata.namespace

https://github.com/pravega/zookeeper-operator/blob/master/charts/zookeeper-operator/templates/operator.yaml#L27-30

The helm chart should provide a way of overriding this so that the operator can watch ALL namespaces (by passing in "") or another namespace different to the one the operator is installed in.

Location
https://github.com/pravega/zookeeper-operator/blob/master/charts/zookeeper-operator/templates/operator.yaml#L27-30

anishakj:

Change log description
Configured WATCH_NAMESPACE field in Helm charts.

Purpose of the change
Fixes #150

What the code does
Added new configuration param in values.yaml file, so that WATCH_NAMESPACE can be configured. Added new cluster roles in zookeeper-operator charts so that it can install zookeeper  in different name space ,other than the one in which operator is running.

How to verify it
Tried installing operator and zk in same and different namespace using charts.


charts/zookeeper-operator/templates/clusterrole.yaml
```diff
@@ -13,15 +13,13 @@ rules:
...
   verbs:
-  - get
-  - watch
-  - list
-- apiGroups:
-  - admissionregistration.k8s.io
-  resources:
   - "*"
-  verbs:
-  - '*'
```
amuraru: @anishakj would it make sense to first integrate https://github.com/pravega/zookeeper-operator/pull/133 and then rebase this patch on top of that?

anishakj:
Closing this PR as invalid, as the changes are messed up

PR #157, opened 2020-03-31
Issue 160: Fix charts lint error and fixed version

kozmagabor:
**Purpose of the change:**
Fixes: #160

**What the code does:**
Updated appVersion and Chart version to latest stable verion

Signed-off-by: Gabor Kozma <gabo@kozmagabor.com>

### DIFF
--- a/charts/zookeeper-operator/Chart.yaml
+++ b/charts/zookeeper-operator/Chart.yaml
@@ -1,6 +1,7 @@
+apiVersion: v1
 name: zookeeper-operator
-version: latest
-appVersion: latest
+version: 0.2.5
+appVersion: 0.2.5
 description: zookeeper operator Helm chart for Kubernetes
 keywords:
 - zookeeper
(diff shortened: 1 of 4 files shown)


spiegela:
@pbelgundi This looks good to me.  I'm fine with pinning our chart to the latest _release_ instead of to the latest tag, but we need to ensure that we bump these files with each subsequent release.

Perhaps we should add a Travis job to check this during PR validation so that we don't miss it?

kozmagabor:
@spiegela @pbelgundi Any updates?

anishakj:
> @spiegela @pbelgundi Any updates?

if we mention stable release instead of latest tag, for  every operator release we have to change the charts. Going forward we can tag latest only for stable releases

pbelgundi:
@anishakj We should be changing the chart version for every release. That is the only way to keep our charts upgradable...

charts/zookeeper-operator/Chart.yaml
```diff
@@ -1,6 +1,7 @@
+apiVersion: v1
 name: zookeeper-operator
-version: latest
-appVersion: latest
+version: 0.2.5
```
pbelgundi, line 3: Latest release version is 0.2.7

kozmagabor: fixed.

kozmagabor:
Thanks!

PR #159, opened 2020-04-02
Added imagePullSecrets to serviceAccount for images in a private registry

azun:
Adding the option `imagePullSecrets` inside the serviceAccount to allow for images in private registries. A list of secret names can be specified in `serviceAccount.imagePullSecrets`

### DIFF
--- a/charts/zookeeper-operator/templates/service_account.yaml
+++ b/charts/zookeeper-operator/templates/service_account.yaml
@@ -1,6 +1,12 @@
 {{- if .Values.serviceAccount.create }}
 apiVersion: v1
 kind: ServiceAccount
+{{- if .Values.serviceAccount.imagePullSecrets }}
+imagePullSecrets:
+{{- range .Values.serviceAccount.imagePullSecrets }}
+  - name: {{ . }}
+{{- end }}
+{{- end }}
 metadata:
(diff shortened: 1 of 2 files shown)


spiegela:
@azun Thanks for this.  I have also seen this field in other charts labeled `registrySecret`.  Could we rename to this for consistency with ad-hoc Helm standards?

Also, this operator is frequently included as a subchart to others, since ZK is often a component of a larger application.  For your use-cases, would it work to nest this value beneath a `global` block like:
```
global:
  registrySecret:  <secret-name>
```

That way this value could be set once in the parent chart?

azun:
Nesting the secrets under global sounds like a good idea. 

I've looked at charts in the official helm/charts github repo and most use a combination of `.Values.global.imagePullSecrets` and/or `.Values.imagePullSecrets`/`.Values.serviceAccount.imagePullSecrets`. I've added global to the chart template, with the option to overwrite it from inside the serviceAccount parameter. 

 I haven't seen `registrySecret` used before and `imagePullSecrets` is more clear in my opinion.

anishakj:
Could you please rebase the changes with master, after that we can merge

azun:
rebased

Issue #162, opened 2020-04-15
Allow a different reference to image in the ZK yaml for Helm


ryphon:
I'm working with Helm and a Jenkins pipeline with some devs that are quite time restrictive.
Our K8s pipeline as is looks for a list of the images to pull and cache so that it doesn't have to do a docker pull for every single environment.  This may be an edge case I should get covered in the pipeline, but for consistency, it seems as if the zookeeper template should be adjusted.

So we end up with a situation where I get this, `Found what appears to be an invalid image name. It will be skipped: ['repository':'private-docker-repo.xyz/zookeeper', 'tag':'latest']` because it's trying to  reconcile the `image: repository: private-docker-repo.xyz tag: latest` as an image to parse, but it's getting a list.

Other helm charts, including the zookeeper-operator, use the ```image: 'private-docker-repo.xyz/zookeeper-operator:latest'``` as one line in the yaml file.

For consistencies sake, I'd like the zk cluster to handle this natively. My current solution is to use the exporter, take the compiled resources and build them into helm chart templates, but seems that I should probably be doing this directly in the crd definition as it gets deployed.

ryphon:
Specifically, it's running helm template and looking for the `image` key and parsing out the value as a string for the docker image to pull.

ryphon:
However, I am seeing that this line... `https://github.com/pravega/zookeeper-operator/blob/6fd02160fbf6a99b515053af47c5958814c23c08/pkg/zk/generators.go#L91` doesn't seem to do what I feel like it's doing..

If I set the image: tag in the yaml to a string, it doesn't seem to properly create the resources as required.

ryphon:
Ah, now I've found `https://github.com/pravega/zookeeper-operator/blob/d84aefddaeb3b457b8478b637406d225330ef38c/pkg/apis/zookeeper/v1beta1/zookeepercluster_types.go#L228` to be where the repository/tag is created via the list in yaml...

Is this something I should just change locally and build the container for my own use case, or do my thoughts about consistency have merit to the overall project?

spiegela:
[...]

PR #163, opened 2020-04-15
Issue 162: Allow image string

ryphon:
Issue: #162 

Purpose is to change from the repository/tag style, to a pure image string, similar to how the operator does it.

Non working, not sure what needs to be changed to allow for 'image' to function as is.

Tests do pass, it does build, just doesn't seem to allow for the image to be passed in like I'd like it to be. Looking for some potential assistance?


anishakj:
Also, could you please share the manifest file that you have used for testing?

ryphon:
Sorry about this, bad github stewardship by me. Closing this pr, the functionality didn't actually work as expected, and I found a way around my initial issue.

PR #168, opened 2020-05-04
added support for current version

Prabhaker24:
Signed-off-by: prabhaker24 <prabhaker.saxena@dell.com>
### Change log description
This PR introduces the CurrentVersion field and sets the CurrentVersion field to the zookeeper cluster image tag deployed by the zookeeper operator 

### Purpose of the change
To display the current version as part of the status field of the zookeeper cluster

### What the code does
It sets the value of the current version in case of the fresh deployment of the zookeeper cluster it also sets the current version when the image tag of the zookeeper cluster is updated and the update is completed

### How to verify it
1) Deploy a Zookeeper cluster and after all the replicas have been deployed the  CurrentVersion 
    field  in the status of zookeeper cluster will show the current image tag deployed by the 
    zookeeper operator.
2) Update  the image tag in the zookeeper cluster and when all the pods have been updated 
    CurrentVersion field will display the new image tag for the zookeeper cluster.

### DIFF
--- a/pkg/apis/zookeeper/v1beta1/zookeepercluster_types.go
+++ b/pkg/apis/zookeeper/v1beta1/zookeepercluster_types.go
@@ -145,7 +145,7 @@ type ZookeeperClusterStatus struct {
 	// CurrentVersion is the current cluster version
 	CurrentVersion string `json:"currentVersion"`
 
-	//UpdateFlag indicate status of update
+	// UpdateFlag indicate if update of the cluster is in progress or not
 	UpdateFlag bool `json:"updateFlag"`
 }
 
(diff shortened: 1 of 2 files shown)


pkg/apis/zookeeper/v1beta1/zookeepercluster_types.go
```diff
@@ -141,6 +141,12 @@ type ZookeeperClusterStatus struct {
 	ExternalClientEndpoint string `json:"externalClientEndpoint"`
 
 	MetaRootCreated bool `json:"metaRootCreated"`
+
+	// CurrentVersion is the current cluster version
+	CurrentVersion string `json:"currentVersion"`
+
+	// UpdateFlag indicate if update of the cluster is in progress or not
+	UpdateFlag bool `json:"updateFlag"`
```
pbelgundi, line 149: Instead of adding a flag it would be better to have `Status Conditions` added to ZK CRD and the Upgrading Condition set to true when Upgrade is in Progress.

Prabhaker24:
Closing this pr as Status condition is added in https://github.com/pravega/zookeeper-operator/pull/176.

PR #179, opened 2020-05-20
Issue 177: Unable to configure storage resources of zookeeper PVCs

### LINKED ISSUE #177: Unable to configure storage resources of zookeeper PVCs
### Description
Unable to configure storage resources of zookeeper PVCs by specifying this value inside the manifest in the following manner at the time of deployment (i.e. the value supplied for this field does not take effect and its value is defaulted to `20Gi` by the operator).

```
  persistence:
    resources:
      requests:
        storage: 10Gi
```

anishakj:
### Change log description
We were not checking whether the persistence volume claim spec has a storage field, before setting the defaults. Due to that configured value was not taking effect

### Purpose of the change
Fixes #177 

### What the code does
Added a check to ensure `storage` field value is empty before setting  deaults 

### How to verify it
Verified by giving different values for storage in manifest file, and pvc are getting created accordingly. If storage is not specified pvcs are created with default size.

### DIFF
--- a/pkg/apis/zookeeper/v1beta1/zookeepercluster_types.go
+++ b/pkg/apis/zookeeper/v1beta1/zookeepercluster_types.go
@@ -34,6 +34,10 @@ const (
 	// container is stopped. This gives clients time to disconnect from a
 	// specific node gracefully.
 	DefaultTerminationGracePeriod = 30
+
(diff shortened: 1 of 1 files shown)


Prabhaker24:
LGTM

PR #194, opened 2020-06-05
Issue 193: Fix envoy waiting block

2phost:
Signed-off-by: Daniel Gonçalves <daniel.dias.g@gmail.com>

The block code that detects and waits for the Envoy does not work because of the ```set -e line```.
The line ```SC=$(curl -s -o /dev/null -w "%{http_code}" http://localhost:15000/ready)``` fails and the script stops running.

I fixed that by disable the flag in this block of code (simple fix).

Edit: format


Prabhaker24:
Thanks for this @2phost, changes LGTM.

PR #199, opened 2020-06-11
Issue 198: Support for name override inside charts

### LINKED ISSUE #198: Adding name overrides for charts
### Description
In the charts we need to give the user the provision of overriding the resource name (which is defaulted to `<release-name>-<chart-name>`) by specifying the desired value for `fullNameOverride` in the values.yaml

### Importance
nice-to-have

### Location
charts

SrishT:

### Change log description
Gives the provision to override the resource name by providing the value for `fullNameOverride` inside values.yaml

### Purpose of the change
Fixes #198 

### What the code does
- If release name is same as the chart name, chart name is used as the resource name.
- If `fullNameOverride` is provided inside the values.yaml file, that value is used as the resource name.

### How to verify it
Tested 3 different scenarios for naming the resources
- `fullNameOverride` is mentioned in values.yaml file (resource name uses the value provided for this field)
- `fullNameOverride` is not mentioned, release name same as chart name (resource name is same as the chart name)
- `fullNameOverride` is not mentioned, release name is different from chart name (resource name is in the given format `<release-name>-<chart-name>`)


anishakj (approved):
LGTM

Issue #118, opened 2020-01-22
Hardcoded cluster domain name


okhotnikov-aa:
Hi,
cluster domain name hardcoded as cluster.local
`func headlessDomain(z *v1beta1.ZookeeperCluster) string {
	return fmt.Sprintf("%s.%s.svc.cluster.local", headlessSvcName(z), z.GetNamespace())
}`
,so you cant resolve names of pods, if your cluster domain name is different.

Actually, looks like svc and cluster domain section may be not neccessary, because pods can resolve names using only service.namespace.

mp-007:
What is the staus here ?

Overwriting domainName should fix the problem ?

`return fmt.Sprintf("%s.%s.svc.%s", headlessSvcName(z), z.GetNamespace(), z.Spec.DomainName)`

anishakj:
Fixed with PR https://github.com/pravega/zookeeper-operator/pull/200

PR #200, opened 2020-06-12
Make kubernetes cluster domain configurable

lukemassa:
### Change log description

Instead of hard-coding "cluster.local" as the kubernetes cluster domain, this makes that setting configurable.

Analogous to this change I made to the Banzai Kafka Operator: https://github.com/banzaicloud/kafka-operator/pull/361

### Purpose of the change

My cluster doesn't use `cluster.local` as our cluster domain, so DNS lookups were failing. Changing it to be configurable allows me to set it in the `ZookeeperCluster` object and successfully bring up a zookeeper cluster.

### What the code does

Adds a parameter to `ZookeeperCluster` called `kubernetesClusterDomain`, which, when set, is used everywhere in the code where `cluster.local` is currently used.

### How to verify it

Kind of tricky if your cluster already works, but you could create a `ZookeeperCluster` with `kubernetesClusterDomain` set to something other than `cluster.local` and watch the cluster fail to create.

I'm not familiar with the Ginkgo testing framework, so I'm not sure if the way I added a test meets its conventions, but it does test for the overridden value.

### DIFF
--- a/pkg/apis/zookeeper/v1beta1/zookeepercluster_types.go
+++ b/pkg/apis/zookeeper/v1beta1/zookeepercluster_types.go
@@ -74,6 +74,9 @@ type ZookeeperClusterSpec struct {
 
 	// Domain Name to be used for DNS
 	DomainName string `json:"domainName,omitempty"`
+
+	// Domain of the kubernetes cluster, defaults to cluster.local
+	KubernetesClusterDomain string `json:"kubernetesClusterDomain,omitempty"`
 }
 
 func (s *ZookeeperClusterSpec) withDefaults(z *ZookeeperCluster) (changed bool) {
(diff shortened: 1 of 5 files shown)


lukemassa:
Whoops, noticed after the fact this addresses: https://github.com/pravega/zookeeper-operator/issues/118

lukemassa:
Reading the linked issue more closely, I'm wondering if this is supposed to be what the role of `DomainName` is? As of right now it seems to be only used for an annotation, if that setting is intended to be used for this purpose, I'm happy to update my code to not include an additional property.

lukemassa:
I don't understand why the coverage went down; I by adding panic()s that I am in fact hitting all of the new lines that I added. However it might be moot depending on the status of `DomainName` vs `KubernetesClusterDomain` so I'll leave it for now

anishakj:
> I don't understand why the coverage went down; I by adding panic()s that I am in fact hitting all of the new lines that I added. However it might be moot depending on the status of `DomainName` vs `KubernetesClusterDomain` so I'll leave it for now

Could you please explain what is the issue you are facing while adding UT?

lukemassa:
Yeah I was just confused because all of the code I introduced was exercised in unit tests, but maybe the issue is like you said it needs to be in zookeepercluster_types_test.go to count, I will look into that.

Also, I opened https://github.com/pravega/zookeeper-operator/issues/208 to address some ambiguity that this PR introduces, not sure what you want to do about that.

lukemassa:
OK added the tests!

anishakj:
Could we make change in helm chart to reflect the new spec parameter `kubernetesClusterDomain` https://github.com/pravega/zookeeper-operator/blob/master/charts/zookeeper/values.yaml

PR #203, opened 2020-06-16
Add support for setting resources for operator pod

### LINKED ISSUE #202: Add support for setting resources for operator pod
### Description

Currently there is no way to set custom pod resources when deploying the operator using helm.

### Importance
Should-have

### Location
helm chart

I'll file a PR

amuraru:
### Purpose of the change

Fixes #202

### What the code does

Add default resources for operator pod and ability to override them.

### How to verify it

_(Steps to verify that the changes are effective)_

### DIFF
--- a/charts/zookeeper-operator/templates/operator.yaml
+++ b/charts/zookeeper-operator/templates/operator.yaml
@@ -34,3 +34,5 @@ spec:
               fieldPath: metadata.name
         - name: OPERATOR_NAME
           value: {{ template "zookeeper-operator.fullname" . }}
+        resources:
+{{ toYaml .Values.resources | indent 10 }}
--- a/charts/zookeeper-operator/values.yaml
+++ b/charts/zookeeper-operator/values.yaml
@@ -23,3 +23,12 @@ crd:
 ## Specifies which namespace the Operator should watch over.
 ## An empty string means all namespaces.
 watchNamespace: ""
+
+## Operator pod resources
+resources: {}
+  # limits:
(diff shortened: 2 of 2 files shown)


charts/zookeeper-operator/templates/operator.yaml
```diff
@@ -34,3 +34,5 @@ spec:
               fieldPath: metadata.name
         - name: OPERATOR_NAME
           value: {{ template "zookeeper-operator.fullname" . }}
+        resources:
+{{ toYaml .Values.resources | nindent 10 }}
```
SrishT, line 38: using `nindent` prints an additional line with `helm template`. Could we use `indent` instead?

charts/zookeeper-operator/values.yaml
```diff
@@ -23,3 +23,13 @@ crd:
...
+
+#
+# Operator pod resources
+resources:
+  limits:
+    cpu: 2
+    memory: 256Mi
+  requests:
+    cpu: 1
+    memory: 128Mi
```
SrishT, line 35: It would be better to comment these values and provide an empty list as default. 
```
resources: {}
  #commented values for limits and requests
```
This way the defaults would be used generally, but can be uncommented if the user wants to set a specific value

charts/zookeeper-operator/templates/operator.yaml
amuraru, line 38:
```suggestion
{{ toYaml .Values.resources | indent 10 }}
```

PR #205, opened 2020-06-17
Issue 204: ZK-operator does not add version to the CR status during upgrade

### LINKED ISSUE #204: ZK-operator does not add version to the CR status during upgrade
### Description

Create zookeeper-operator and zk cluster. Upgrade the operator from `0.2.7` latest master image that is built. In zookeeper cluster status version is not getting updated.

### Importance

Must have

### Location

https://github.com/pravega/zookeeper-operator/blob/master/pkg/apis/zookeeper/v1beta1/zookeepercluster_types.go

This issue is introduced, because we have added metrics ports for  monitoring. If in the prev version metrics port is not there, while doing upgrade metrics port is not set to default port.

anishakj:

### Change log description
With recent changes we have added metrics ports. While doing upgrade, metrics port is not set to a default value because in `withDefaults` we are only checking ports are not present. we were not checking for individual ports for client, quorum, leader-election and metrics are present. Due to this reconcile was failing.
### Purpose of the change

 Fixes #204

### What the code does

In case we have not specified all the ports in zookeeper cluster manifest,  in `withDefaults`  method, we will set the missing ports to default.

### How to verify it

1. Verified zookeeper cluster status is set properly in case of upgrade
2. Verified by not giving client ports in the zk manifest, and client port is setting to default.

### DIFF
--- a/pkg/apis/zookeeper/v1beta1/zookeepercluster_types.go
+++ b/pkg/apis/zookeeper/v1beta1/zookeepercluster_types.go
@@ -106,7 +106,9 @@ func (s *ZookeeperClusterSpec) withDefaults(z *ZookeeperCluster) (changed bool) 
 		}
 		changed = true
 	} else {
-		var foundClient, foundQuorum, foundLeader, foundMetrics bool
+		var (
+			foundClient, foundQuorum, foundLeader, foundMetrics bool
+		)
 		for i := 0; i < len(s.Ports); i++ {
 			if s.Ports[i].Name == "client" {
 				foundClient = true
--- a/pkg/apis/zookeeper/v1beta1/zookeepercluster_types.go
+++ b/pkg/apis/zookeeper/v1beta1/zookeepercluster_types.go
@@ -105,7 +105,41 @@ func (s *ZookeeperClusterSpec) withDefaults(z *ZookeeperCluster) (changed bool) 
 			},
 		}
 		changed = true
+	} else {
+		var foundClient, foundQuorum, foundLeader, foundMetrics bool
+		for i := 0; i < len(s.Ports); i++ {
+			if s.Ports[i].Name == "client" {
+				foundClient = true
+			} else if s.Ports[i].Name == "quorum" {
+				foundQuorum = true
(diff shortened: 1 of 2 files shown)


pkg/apis/zookeeper/v1beta1/zookeepercluster_types.go
```diff
@@ -105,7 +105,41 @@ func (s *ZookeeperClusterSpec) withDefaults(z *ZookeeperCluster) (changed bool)
 			},
 		}
 		changed = true
+	} else {
+		var foundClient, foundQuorum, foundLeader, foundMetrics bool
```
Prabhaker24, line 109: we can create a variable block here.

pkg/apis/zookeeper/v1beta1/zookeepercluster_types_test.go
```diff
@@ -241,6 +242,42 @@ var _ = Describe("ZookeeperCluster Types", func() {
...
+					Name: "example",
+				},
+				Spec: v1beta1.ZookeeperClusterSpec{
+					Ports: ports,
+				},
+			}
+			z.WithDefaults()
+			p = z.ZookeeperPorts()
+		})
+		It("should have a client port", func() {
```
Prabhaker24, line 269: we can remove this assertion as we are only setting it and it's no set by the "withDefaults()" function.

pkg/apis/zookeeper/v1beta1/zookeepercluster_types.go
anishakj: Done

pkg/apis/zookeeper/v1beta1/zookeepercluster_types_test.go
anishakj: we are creating a dummy value and setting defaults so that all the conditions in else is getting executed.

Prabhaker24: okay

Prabhaker24:
LGTM

PR #221, opened 2020-07-15
ZooKeeper Operator API Resource Hook.

co-jo:
### Change log description

* Adds a manifest which defines a post-install/upgrade hook to make sure the `zookeepercluster` resource exists.

### Purpose of the change

* If one is unlucky with the timings when install the `zookeeper-operator` chart, directly followed by the `zookeeper` chart, the api-resource `ZooKeeperCluster` may not exist by the time the zookeeper chart tries to create a `ZooKeeperCluster`.

### What the code does

* See above.

### How to verify it

Run many install/uninstall cycles of the mentioned charts -- see that the zookeeper chart does not start before its resource dependency is registered.

### DIFF
--- a/charts/zookeeper-operator/templates/post-install-upgrade-hooks.yaml
+++ b/charts/zookeeper-operator/templates/post-install-upgrade-hooks.yaml
@@ -69,7 +69,7 @@ data:
     set -e
     sleep 30
 
-    if [ -z "$(kubectl api-resources | grep zk)" ]; then
+    if [ -z "$(kubectl api-resources | grep ZookeeperCluster)" ]; then
         exit 1
     fi
 ---
@@ -84,7 +84,7 @@ metadata:
     "helm.sh/hook-weight": "2"
     "helm.sh/hook-delete-policy": hook-succeeded, before-hook-creation, hook-failed
 spec:
-  backoffLimit: 3
+  backoffLimit: {{ .Values.hooks.backoffLimit }}
   template:
     metadata:
       name: {{ template "zookeeper-operator.fullname" . }}-post-install-upgrade
--- a/charts/zookeeper-operator/templates/post-install-upgrade-hooks.yaml
+++ b/charts/zookeeper-operator/templates/post-install-upgrade-hooks.yaml
@@ -73,16 +73,12 @@ data:
         exit 1
     fi
 
-    replicas=`kubectl get deployment {{ template "zookeeper-operator.fullname" . }} \
-      -n {{ .Release.Namespace }} \
-      -o jsonpath='{.status.replicas}'`
     readyReplicas=`kubectl get deployment {{ template "zookeeper-operator.fullname" . }} \
       -n {{ .Release.Namespace }} \
       -o jsonpath='{.status.readyReplicas}'`
-    echo "ZookeeperCluster replicas: $replicas"
     echo "ZookeeperCluster readyReplicas: $readyReplicas"
 
-    if [ $readyReplicas != $replicas ]; then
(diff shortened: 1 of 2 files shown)


co-jo:
Note: I'm not entirely sure if all the resources mentioned in the `yaml` manifest area needed. I referenced the one from the zookeeper chart.

charts/zookeeper-operator/templates/post-install-upgrade-hooks.yaml
```diff
@@ -0,0 +1,116 @@
...
+  validations.sh: |
+    #!/bin/sh
+    set -e
+    sleep 30
+
+    if [ -z "$(kubectl api-resources | grep zk)" ]; then
+        exit 1
+    fi
+
+    readyReplicas=`kubectl get deployment {{ template "zookeeper-operator.fullname" . }} \
```
SrishT, line 76: If the purpose of the hook is to wait for the deployment to become ready then that can be achieved by adding `--wait` with the `helm install` command.
I thought the purpose of the hook was to check that the `api-resource` for zookeeperclusters have been successfully created or not. So the operator deployment should succeed only if the api-resource is successfully created, and should fail if not, and this PR does not do that.

charts/zookeeper-operator/templates/post-install-upgrade-hooks.yaml
```diff
@@ -0,0 +1,107 @@
...
+      "helm.sh/hook": post-install, post-upgrade
+      "helm.sh/hook-weight": "1"
+      "helm.sh/hook-delete-policy": hook-succeeded, before-hook-creation, hook-failed
+data:
+  validations.sh: |
+    #!/bin/sh
+    set -e
+    sleep 30
+
+    if [ -z "$(kubectl api-resources | grep zk)" ]; then
```
SrishT, line 72: This works too, but just for the sake of consistency can u replace `zk` with `ZookeeperCluster` ?

charts/zookeeper-operator/templates/post-install-upgrade-hooks.yaml
```diff
@@ -0,0 +1,107 @@
...
+kind: Job
+metadata:
+  name: {{ template "zookeeper-operator.fullname" . }}-post-install-upgrade
+  namespace: {{ .Release.Namespace }}
+  annotations:
+    "helm.sh/hook": post-install, post-upgrade
+    "helm.sh/hook-weight": "2"
+    "helm.sh/hook-delete-policy": hook-succeeded, before-hook-creation, hook-failed
+spec:
+  backoffLimit: 3
```
SrishT, line 87: Could u use the backOffLimit value that u have taken as input from the user in values.yaml instead of hardcoding it?

charts/zookeeper-operator/values.yaml
```diff
@@ -39,3 +39,9 @@ resources: {}
   # requests:
   #   cpu: 1
   #   memory: 128Mi
+
+hooks:
+  image:
+    repository: lachlanevenson/k8s-kubectl
+    tag: v1.16.10
+
```
SrishT, line 47: U could accept the backOffLimit here like u did for the other operators

charts/zookeeper-operator/templates/post-install-upgrade-hooks.yaml
co-jo: The reason I passed in `zk` instead of  `ZookeeperCluster` was to be consistent. I notice that in hooks manifest for the `zookeeper` chart it referenced `zk`.

PR #227, opened 2020-08-12
Adopt controller-gen to generate crd manifests and adopt helm v3 way of managing crds

jiayushe:
### Change log description

Adopt `controller-gen` to generate crd manifests and adopt helm v3 way of managing crds

### Purpose of the change

- Clean up management of crd manifests
- Prevent inconsistency of crd manifests in manual and helm deployments

### What the code does

- Fix minor typo in `README` and `Makefile`
- Fix outdated `pkg/apis/zookeeper/v1beta1/zz_generated.deepcopy.go` using `operator-sdk generate k8s`
- Adopt `controller-gen` to generate crd manifests, this prevents inconsistency of crd manifests in manual and helm deployments
- Adopt helm v3 way of managing crds (https://helm.sh/docs/topics/charts/#custom-resource-definitions-crds)

### How to verify it

Run `make manifests` to generate crd manifests in `deploy/crds` and `charts/zookeeper-operator/crds`


anishakj:
In the crd  generated, all the optional params of the spec are made as required fields

jiayushe:
Hi @anishakj , thanks for reviewing. Whether a param is optional or required is inferred from the JSON tag in the definition. Since they are optional, I have tagged them as optional in the main code. The generated CRD will not mark them as required now. The test should be passing as well.

pkg/apis/zookeeper/v1beta1/zookeepercluster_types.go
```diff
@@ -188,8 +188,19 @@ func (s *ZookeeperClusterSpec) withDefaults(z *ZookeeperCluster) (changed bool)
 }
 
 // +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
-// ZookeeperCluster is the Schema for the zookeeperclusters API
 // +k8s:openapi-gen=true
```
Prabhaker24, line 191: can we add a comment here specifying below makers comments are for generating crd using kubebuilder ( as we use operator-sdk framework for other things I feel it would be better to mention that)

Prabhaker24 (approved):
other than this minor comment this pr LGTM, Thanks @jiayushe for doing this.

pkg/apis/zookeeper/v1beta1/zookeepercluster_types.go
jiayushe: Done. Thanks for reviewing.

charts/zookeeper-operator/templates/crd.yaml
```diff
@@ -1,47 +1,6 @@
 {{- if .Values.crd.create }}
```
SrishT, line 1: This file addition is unnecessary. I feel u should revert this to your previous change and remove the option for user to specify whether the crd needs to be created from the values.yaml. As per my understanding from [this](https://github.com/helm/helm/pull/6243) PR :
```
If a CRD is already found in the cluster, Helm will skip installing that CRD and will log this to the debug log. It will neither error out nor see if there are any differences between its version and the existing version.
```

SrishT (changes requested):
Also remove the following section from the values.yaml section of the zookeeper-operator charts, as you had previously done.
```
crd:	
  create: true
```

deploy/crds/zookeeper.pravega.io_zookeeperclusters.yaml
```diff
@@ -0,0 +1,1338 @@
+
+---
```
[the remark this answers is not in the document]
jiayushe: The separator is autogenerated. I don't think there is a way to remove it except manually, which would make it harder to maintain.

charts/zookeeper-operator/templates/crd.yaml
jiayushe: Okay I will remove it then. I was thinking that this could be used for backward compatibility with helm v2.

SrishT: We have migrated to helm 3 now, and don't plan to support helm 2

anishakj (approved):
LGTM

deploy/crds/zookeeper.pravega.io_zookeeperclusters.yaml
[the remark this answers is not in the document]
SrishT: okay

SrishT (approved):
LGTM

PR #231, opened 2020-08-18
Issue 222: zoo.cfg updated parameters are not picked up during rolling restarts

### LINKED ISSUE #222: zoo.cfg updated parameters are not picked up during rolling restarts
### Description

Updating ZK parameters like `initLimit`, `tickTime` are not reflected in ZK when the cluster is restarted. 
Steps to reproduce:
- deploy a new ZK cluster 
- update `initLimit` in ZK spec
- rolling restart is triggered and one would expect the new value to be reflected in zk runtime
This is not happing and `/data/conf/zoo.cfg` configuration file used in zk command line does not reflect the change.

https://github.com/adobe/zookeeper-operator/blob/dede2bacf9ff56fa85bd81286cea0fd20815e632/pkg/zk/generators.go#L203-L217

#### Analysis

Following the param change it seems the `/conf/zoo.cfg` correctly picks up the change but the actual `/data/conf/zoo.cfg` is only copied once during initial installation (and never updated)
https://github.com/adobe/zookeeper-operator/blob/master/docker/bin/zookeeperStart.sh#L140-L143

### Importance
must-have

anishakj:

### Change log description

We were copying zoo.cfg file from /conf directory to /data/conf.cfg only at the start. If any configuration parameters are changed after installation, it was not taking into effect.

### Purpose of the change

 Fixes #222

### What the code does

Modified the start up script such a way that /conf/zoo.cfg contents are copied to  /data/conf.cfg even on restart. Will retain the dynamicconfig file in /data/conf.cfg

### How to verify it

Verified by changing the config parameters and saw that they are getting updated on rolling restarts.

### DIFF
--- a/docker/bin/zookeeperStart.sh
+++ b/docker/bin/zookeeperStart.sh
@@ -144,14 +144,17 @@ fi
 
 ZOOCFGDIR=/data/conf
 export ZOOCFGDIR
+echo Copying /conf contents to writable directory, to support Zookeeper dynamic reconfiguration
 if [[ ! -d "$ZOOCFGDIR" ]]; then
-  echo Copying /conf contents to writable directory, to support Zookeeper dynamic reconfiguration
   mkdir $ZOOCFGDIR
   cp -f /conf/zoo.cfg $ZOOCFGDIR
-  cp -f /conf/log4j.properties $ZOOCFGDIR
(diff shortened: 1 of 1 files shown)


Prabhaker24 (approved):
LGTM

Issue #208, opened 2020-06-19
Clarify difference between KubernetesClusterDomain and DomainName


lukemassa:
### Description

I recognize first of all this is a problem entirely of my own creation. When adding https://github.com/pravega/zookeeper-operator/pull/200 I created the notion of KubernetesClusterDomain (i.e. defaulting to cluster.local, but overideable, corresponding to https://github.com/kubernetes/dns/blob/master/cmd/kube-dns/app/options/options.go#L57, described here: https://kubernetes.io/docs/tasks/administer-cluster/dns-custom-nameservers/#introduction).

However, there also exists a setting `DomainName` (https://github.com/pravega/zookeeper-operator/blob/master/pkg/apis/zookeeper/v1beta1/zookeepercluster_types.go#L76), whose only purpose seems to be setting the `external-dns.alpha.kubernetes.io/hostname` on the Service (described here: https://github.com/kubernetes-sigs/external-dns/blob/master/docs/faq.md).

My question is whether we should have related the two at all? If the idea is that `DomainName` is how this service is addressed from outside the cluster, completely out of the control of the operator, whereas KubernetesClusterDomain helps build the internal DNS name, maybe we should change the name of the former to `ExternalHostName` and add a few lines in the comments around the setting to clarify their different usage?

Even if it's decided it's not worth changing the variable name, given that my understanding is correct I'd be happy to write up the comments to clarify the role of the settings.

### Importance

Minimal, documentation/clarification.

### Location

`pkg/apis/zookeeper/v1beta1/zookeepercluster_types.go`

### Suggestions for an improvement

Given that my understanding above is correct, comments to clarify the roles and/or changing the variable name to make it clear it's external.

If my understanding is wrong, and these values are related somehow, my suggestion would be one of them be derivable from the other.

anishakj:
@lukemassa  your understanding is correct. DomainName in the spec is used for setting the annotations.

anishakj:
We wont be able to change the spec parameter as it can cause compatibility  issues while doing upgrade.

Prabhaker24:
I have raised a pr and have changed the description of the DomainName field to reflect what it does we can't change the field now as @anishakj  has rightly pointed out that it will cause compatibility issues during our operator upgrades.

Prabhaker24:
closing this issue as the pr https://github.com/pravega/zookeeper-operator/pull/234 is merged.

PR #234, opened 2020-08-27
Issue 233: Fixed chart bug and made ready and unready fields as nullable and added checks for various crd fields

### LINKED ISSUE #233: Create zookeeper cluster is failing with chart due to domain name not specified correctly in charts
### Description
Create zookeeper cluster is failing with chart due to domain name not specified correctly in charts

### Importance
must_have

### Location
charts/zookeeper

add a check in zookeeper.yaml file in chart so that domainname field does not get null as value in case domainname value inside the values.yaml file is empty.

Prabhaker24:
Signed-off-by: prabhaker24 <prabhaker.saxena@dell.com>

### Change log description
Create zookeeper cluster is failing with chart due to domain name not specified correctly in charts Ater the OpenApi SchemaValidation check Added and made the ready and unready member list as nullable in crd.

### Purpose of the change
Fixes #233 #208 

### What the code does
Because of the OpenApiSchemaValidation being present we can't have the domainname with null value which was the case till now so I have changed it so that if no value is provided domainname field get's skipped along with that to keep backwards compatibility that is for the latest operator code to be able to deploy zookeeper cluster 0.2.8 or before we need to make ready and unready member list as nullable in the crd as they were not omitempty before. 

Additionally, I have removed controller-gen command from make file and have used operator-sdk command to create the crd below is the command:-

``` operator-sdk generate crds ```

And have added validation checks for various fields in the crd.

### How to verify it
Create zookeeper cluster using charts after deploying the zookeeper-operator, Cluster should get deployed.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -71,7 +71,7 @@ test-e2e-remote: test-login
 	docker push $(TEST_IMAGE)
 	operator-sdk test local ./test/e2e --operator-namespace default \
 		--namespaced-manifest ./test/e2e/resources/rbac-operator.yaml \
-		--global-manifest deploy/crds/zookeeper.pravega.io_zookeeperclusters.yaml \
+		--global-manifest deploy/crds/zookeeper.pravega.io_zookeeperclusters_crd.yaml \
 		--image $(TEST_IMAGE) --go-test-flags "-v -timeout 0"
 
 test-e2e-local:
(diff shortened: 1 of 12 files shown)


anishakj (changes requested):
Please add validation spec fields like pullpolicy, container port etc. And test the negative scenarios and ensure proper error is given

anishakj:
Could you please check can we add validations for storagetype  and access modes

anishakj (approved):
LGTM

PR #243, opened 2020-09-16
Use getent instead of nslookup for starting scripts

jinfwhuang:
The nslookup only exit with status 0 with successful lookup of dns record for both forward and backward lookups. It would mean that the cluster's DNS setup has to support. `getent` only performs a forward lookup.

pravega#76

### DIFF
--- a/docker/bin/zookeeperFunctions.sh
+++ b/docker/bin/zookeeperFunctions.sh
@@ -19,8 +19,8 @@ function zkConfig() {
 function zkConnectionString() {
   # If the client service address is not yet available, then return localhost
   set +e
-  nslookup "${CLIENT_HOST}" 2>/dev/null 1>/dev/null
-  if [[ $? -eq 1 ]]; then
+  getent hosts "${CLIENT_HOST}" 2>/dev/null 1>/dev/null
+  if [[ $? -ne 0 ]]; then
     set -e
     echo "localhost:${CLIENT_PORT}"
   else
--- a/docker/bin/zookeeperReady.sh
+++ b/docker/bin/zookeeperReady.sh
@@ -25,8 +25,8 @@ OK=$(echo ruok | nc 127.0.0.1 $CLIENT_PORT)
 # Check to see if zookeeper service answers
 if [[ "$OK" == "imok" ]]; then
   set +e
-  nslookup $DOMAIN
-  if [[ $? -eq 1 ]]; then
+  getent hosts $DOMAIN
+  if [[ $? -ne 0 ]]; then
     set -e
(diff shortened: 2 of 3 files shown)


amuraru:
@jinfwhuang can you please rebase your patch on top of latest master? thanks

jinfwhuang:
@amuraru  Updated this PR.

derekm:
I think the claim that `nslookup` requires both forward A/AAAA records and reverse PTR records is false.

Rather, `nslookup` uses DNS-only lookups and does not return entries from `/etc/hosts`. Whereas, `getent hosts` does resolution according to Name Service Switch configuration.

This being the case, using `nslookup` does hinder `/etc/hosts`-only deployments.

amuraru:
@derekm thanks for your analysis. 
I don't have a strong opinion for this patch - we've seen this on some of our clusters but those are not critical atm so we can drop this.

jinfwhuang:
I think getent is simpler than nslookup. It works for more cases. Why not?

anishakj:
@jinfwhuang Are you still facing issues with nslookup?

jinfwhuang:
> @jinfwhuang Are you still facing issues with nslookup?

We have been running our zookeeper clusters with the patch in the past year.

anishakj:
> > @jinfwhuang Are you still facing issues with nslookup?
> 
> We have been running our zookeeper clusters with the patch in the past year.

@jinfwhuang Could you please sign-off the commit?

jinfwhuang (commented):
Looks good to me

anishakj:
> Looks good to me

@jinfwhuang If you see DCO check is failing, because commits are not signed off

jinfwhuang:
> > Looks good to me
> 
> @jinfwhuang If you see DCO check is failing, because commits are not signed off

Okay. Updated the commit.

anishakj:
> > > Looks good to me
> > 
> > 
> > @jinfwhuang If you see DCO check is failing, because commits are not signed off
> 
> Okay. Updated the commit.
> 
> Question, why was signoff necessary for this project?

In Pravega, we are using DCO  instead of CLA, so authors acknowledge and retain their copyrights

anishakj (approved):
LGTM

derekm (approved):
It will be nice to help a user get off of a private fork!

amuraru:
@jinfwhuang I still see a reference to nslookup here https://github.com/pravega/zookeeper-operator/blob/c4113df5c90e290aebe2007041c188959ddb2fb4/docker/bin/zookeeperStart.sh#L77

Should this be updated as well?
thanks

PR #246, opened 2020-09-21
Issue 240: Bootstrap Process does not handle missing DNS server correctly

### LINKED ISSUE #240: Bootstrap Process does not handle missing DNS server correctly
### Description
Zookeeper Node bootstrap does not handle DNS not being ready and incorrectly takes a missing DNS server as meaning there are no existing ZK nodes.  This results in it incorrectly setting up a new bootstrap node and causing a "split-brain" situation on the cluster where the new node has no data from the existing ZK cluster.

### Importance
Critical as this can cause production outages when systems such as Flink attempt to recover Jobs based on information in the ZK node.

### Location
During start up of a Zookeeper node that has no existing data the `zookeeperFunctions.sh` attempts an `nslookup` and will take any error response to mean that there are no existing nodes.  However in rare node recovery type situations, the DNS servers my not be ready when the ZK POD checks and instead gets:
```
+ nslookup zookeeper-headless.edge.svc.cluster.local
;; connection timed out; no servers could be reached
```
Unfortunately `zookeeperFunctions.sh` takes this as an error condition that there are no existing nodes and bootstraps a new cluster.

It should somehow identify when `nslookup` has been successful but has no servers Vs when nslookup has failed to contact a DNS server.

anishakj:

### Change log description

Fix is added to parse nslookup  output to differentiate lookup failure and DNS failure.

### Purpose of the change

Fixes #240

### What the code does

In some scenarios, like DNS servers are not ready , nslookup  was fialing. In the code we were handling it for all the nodes except the first node. In case of DR , the node that is shutdown is the node containing `zookeeper-0`,  nslookup failure is causing split brain situation and it is creating a new cluster. 

Fix is added to parse the nslookup output and differentiate DNS failure vs lookup failure.

### How to verify it
Verified, restart of of zookeeper pods, scale up and down.
In baremetal cluster deleted node containing zookeeper-0.

### DIFF
--- a/docker/bin/zookeeperStart.sh
+++ b/docker/bin/zookeeperStart.sh
@@ -93,7 +93,6 @@ else
       break
     fi
   done
-
 fi
 
 if [[ "$ONDISK_MYID_CONFIG" == true && "$ONDISK_DYN_CONFIG" == true ]]; then
--- a/docker/bin/zookeeperStart.sh
+++ b/docker/bin/zookeeperStart.sh
@@ -72,13 +72,18 @@ set -e
 # Determine if there is a ensemble available to join by checking the service domain
 set +e
 nslookup $DOMAIN
-if [[ $? -eq 1 ]]; then
+if [[ $? -eq 0 ]]; then
+  ACTIVE_ENSEMBLE=true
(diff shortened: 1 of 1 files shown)


Prabhaker24 (approved):
LGTM

PR #253, opened 2020-10-06
Issue 250: Not specifying a default storageclass

### LINKED ISSUE #250: standard should not be set as the default storageclass
### Description
Currently the default storageclass is being set to standard which might not work for all kinds of environments. Hence we should let Kubernetes pick the default storageclass instead of explicitly specifying it. For more details refer to [this](https://kubernetes.io/docs/reference/access-authn-authz/admission-controllers/#defaultstorageclass).

### Importance
should have 

### Location
charts

`standard` is the default storageclass in PKS and a few more environments but not all. By not specifying anything for the storageclass, the default storageclass for that particular environment will be picked.

SrishT:

### Change log description
Does not specify any default value for storageclass which allows the default storageclass for the given environment to be picked instead of having to hard-code it.

### Purpose of the change
Fixes #250 

### How to verify it
Install the zookeeper charts which do not specify any value for the storageclass. The following command shows that the default storageclass for the environment will be automatically picked.
```
$ kubectl get storageclass
NAME                 PROVISIONER                                       AGE
nfs                  cluster.local/nfs-server-provisioner-1599805455   25d
standard (default)   kubernetes.io/vsphere-volume                      25d

$ kubectl get pvc
NAME                    STATUS   VOLUME                                     CAPACITY   ACCESS MODES   STORAGECLASS   AGE
data-zookeeper-0        Bound    pvc-6120fb84-919b-402b-812b-834b79edc649   20Gi       RWO            standard       69m
data-zookeeper-1        Bound    pvc-0de5bc53-e6a5-40ba-8b8e-d7d65a1bcc85   20Gi       RWO            standard       68m
data-zookeeper-2        Bound    pvc-0b6d6218-7c78-4496-b711-f27d006c88aa   20Gi       RWO            standard       68m

```

### DIFF
--- a/charts/zookeeper/README.md
+++ b/charts/zookeeper/README.md
@@ -77,7 +77,7 @@ The following table lists the configurable parameters of the zookeeper chart and
 | `config.quorumListenOnAllIPs` | Whether Zookeeper server will listen for connections from its peers on all available IP addresses | `false` |
 | `storageType` | Type of storage that can be used it can take either ephemeral or persistence as value | `persistence` |
 | `persistence.reclaimPolicy` | Reclaim policy for persistent volumes | `Delete` |
-| `persistence.storageClassName` | Storage class for persistent volumes | `standard` |
+| `persistence.storageClassName` | Storage class for persistent volumes | `` |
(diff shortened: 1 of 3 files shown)


anishakj (approved):
LGTM

PR #255, opened 2020-10-08
Issue78: Allow to set a SeviceAccount to cluster Statefulset 

### LINKED ISSUE #78: allow to set a SeviceAccount to cluster Statefulset
We need to be able to define a Kubernetes `ServiceAccount` in the Statefulset, so we can use images from private repositories (with specific ImagePullSecrets).

anishakj:
### Change log description

Added a new field service account in podspec so that users will be able to create pods with  non default service accounts

### Purpose of the change

 Fixes #78

### What the code does
Added service account to pod spec. And added corresponding changes in zookeeper cluster charts.
### How to verify it

Verified the zookeeper operator and cluster installation
Verified zookeeper operator and cluster upgrades

### DIFF
--- a/charts/zookeeper/templates/role.yaml
+++ b/charts/zookeeper/templates/role.yaml
@@ -1,7 +1,7 @@
 kind: Role
 apiVersion: rbac.authorization.k8s.io/v1
 metadata:
-  name:{{ default "zookeeper" .Values.pod.serviceAccountName }}
+  name: {{ default "zookeeper" .Values.pod.serviceAccountName }}
   namespace: {{ .Release.Namespace }}
   labels:
 {{ include "zookeeper.commonLabels" . | indent 4 }}
--- a/charts/zookeeper/templates/role.yaml
+++ b/charts/zookeeper/templates/role.yaml
@@ -1,7 +1,7 @@
 kind: Role
 apiVersion: rbac.authorization.k8s.io/v1
 metadata:
-  name: {{ .Values.serviceAccount.name }}
+  name:{{ default "zookeeper" .Values.pod.serviceAccountName }}
   namespace: {{ .Release.Namespace }}
   labels:
 {{ include "zookeeper.commonLabels" . | indent 4 }}
--- a/charts/zookeeper/templates/role.yaml
+++ b/charts/zookeeper/templates/role.yaml
@@ -1,0 +1,21 @@
+kind: Role
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: {{ .Values.serviceAccount.name }}
+  namespace: {{ .Release.Namespace }}
+  labels:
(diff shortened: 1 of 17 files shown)


charts/zookeeper/templates/role.yaml
```diff
@@ -0,0 +1,21 @@
...
+metadata:
+  name: {{ default "zookeeper" .Values.pod.serviceAccountName }}
+  namespace: {{ .Release.Namespace }}
+  labels:
+{{ include "zookeeper.commonLabels" . | indent 4 }}
+rules:
+- apiGroups:
+  - zookeeper.pravega.io
+  resources:
+  - "*"
```
gaddas3, line 12: shall we specify resources instead of *

Prabhaker24 (approved):
LGTM

SrishT (approved):
LGTM

charts/zookeeper/templates/role.yaml
anishakj: I  am not sure what resources we can give here. In our other operators as well, we have given `*`

gaddas3: keep investigating what resources are needed exactly. we can move forward with current PR.

gaddas3 (approved):

charts/zookeeper/templates/role.yaml
maddisondavid: It seems valid to me to be `*` since these are Zookeeper CRD resources which the Zookeeper operator is responsible for.

sarlaccpit: Why is the service account that is used for the zk cluster pods given access to zookeeper operator crd resources?

If this role was given to the zookeeper-operator service account I'd understand but... actually it already has what it needs: https://github.com/pravega/zookeeper-operator/blob/master/charts/zookeeper-operator/templates/clusterrole.yaml

I thought the point of this PR was to give the ability for zk pods to run a non default service account (which is achieved as part of the changes in zookeepercluster_types.go ); but I don't understand why this service account needs to be given any role.   Think of the roles the "default" service account has when it runs the zk pods before this PR.  None (explicitly).

sarlaccpit: So in summary: I understand and agree with all the changes in this PR except I don't understand any of the changes in charts/zookeeper/templates, except for service_account.yaml and zookeeper.yaml

sarlaccpit: Additionally, I'd suggest that while the service account name should come from values.yaml, really the zk operator would actually provision the Service account resource itself (if name is provided).

anishakj: Thanks @sarlaccpit for your review. I have addressed your comments in another PR https://github.com/pravega/zookeeper-operator/pull/256

Issue #257, opened 2020-10-12
ZookeeperCluster spec.labels are not propagated to StatefuleSet/Pods created by the operator


amuraru:
### Description

In our environment we need to have custom labels attached to resources created by the zookeeper-operator for internal compliance.
Trying to use [`spec.labels`](https://github.com/pravega/zookeeper-operator/blob/master/deploy/crds/zookeeper.pravega.io_zookeeperclusters_crd.yaml#L1197-L1202) in the `ZookeeperCluster` CRD doesn't seem to work per documentation. The labels are correctly added to ZookeeperCluster CR but not propaged to STS or Pods.

### Suggestions for an improvement

Propaged ZookeeperCluster CR labels to resources managed by the operator

PR #258, opened 2020-10-12
ISSUE-257 Propagate labels to ZookeeperCluster underlying resources

amuraru:
Fixes: #257


spiegela (commented):
This looks great. I was needing this exact fix myself. Could you also add unit tests for these labels?

amuraru:
added unit tests

spiegela (approved):
💯 Perfect.  Thanks!

anishakj (approved):
LGTM

endzyme:
I think this may have broken somewhere along the way. I am unable to propagate labels down to the statefulset nor the pods. I will open an issue for this.

endzyme:
Ah - the problem is that labels from `spec.labels` are propagated to pods, but not `spec.pod.labels`

PR #261, opened 2020-10-20
Fix marker comments for runtime object deepcopy

jiayushe:
### Change log description

Fix marker comments for runtime object deepcopy.

### Purpose of the change

The marker comments for runtime object deepcopy was wrongly located. As such, the runtime object deepcopy code was not generated via `operator-sdk generate k8s` as expected. This change fixes the issue.

### What the code does

Fix location of marker comments for runtime object deepcopy so that the deepcopy code gets generated correctly.

### How to verify it

`operator-sdk generate k8s` should generate runtime object deepcopy code correctly.


anishakj (approved):
LGTM

PR #271, opened 2020-11-12
Issue 165: Adding annotations to PVC

### LINKED ISSUE #165: Can we add annotations for the pvc created?
I know we can pass the `storageClassName` in the `pvc` spec and we are using that right now. But **is it possible** to pass the annotations for the `pvc` template?

It would be really helpful if we could do that as we use portworx for our pvc and we need to add some annotations for it to work, specifically these annotations are storage class and the secret name used for encryption.

If this is not supported, I'd be happy to help in adding this if someone can point me in the right direction.

anishakj:

### Change log description
Added a new field annotations to persistence and populating while pvc is created

### Purpose of the change

 Fixes #165 

### What the code does

Provided support for adding annotations to PVCs
### How to verify it

Created zookeeper cluster, by specifying annotations in values.yaml file. Annotations are populated for PVC created

### DIFF
--- a/charts/zookeeper-operator/templates/zookeeper.pravega.io_zookeeperclusters_crd.yaml
+++ b/charts/zookeeper-operator/templates/zookeeper.pravega.io_zookeeperclusters_crd.yaml
@@ -1271,6 +1271,12 @@ spec:
                   - Delete
                   - Retain
                   type: string
+                annotations:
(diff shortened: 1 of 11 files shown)


Prabhaker24 (approved):
LGTM

SrishT (approved):
LGTM

PR #282, opened 2020-12-09
Add support to add annotations to operator pod

amuraru:
### Change log description

Add support to add annotations to operator pod in helm chart

### Purpose of the change

Ability to set custom annotations to pods

### What the code does

Extends helm chart


SrishT (approved):
LGTM

PR #290, opened 2021-01-11
Issue 275: Made changes for configuring probes

### LINKED ISSUE #275: make LivenessProbe,ReadinessProbe initialdelayseconds,timeout configurable thru CRD
### Description

_(Describe the feature, bug, question, proposal that you are requesting)_
In cases where ZKserver takes little bit more time to startup,  "echo ruok | nc 127.0.0.1 $CLIENT_PORT" fails with connection refused. ZookeeperReady.sh doesn't promote the node as participant and the quorum doesn't form
### Importance

_(Indicate the importance of this issue to you (blocker, must-have, should-have, nice-to-have))_
blocker as the  zk-2/zk-3 pod goes in to error state, doesn't get promoted as participant and the quorum is not formed
### Location

ZK cluster CRD POD configuration

please make the initialdelayseconds configurable through CRD so some delay can be introduced 

anishakj:

### Change log description

Made the changes for configuring readiness probe and liveness probe timeout values

### Purpose of the change

 Fixes #275

### What the code does

User will be able to specify initial delay seconds, timeout, success and failure threshold via charts
### How to verify it

Verified that able to configure values for readiness and liveness probes


charts/zookeeper/values.yaml
```diff
@@ -1,15 +1,26 @@
 replicas: 3
 
 image:
-  repository: pravega/zookeeper
-  tag: 0.2.9
+  repository: devops-repo.isus.emc.com:8116/nautilus/zookeeper
```
Prabhaker24, line 4: Looks like you changed this by mistake, Please revert this change.

charts/zookeeper/values.yaml
```diff
@@ -1,15 +1,26 @@
...
+  tag: 0.2.9-162-9bdc9a2
```
Prabhaker24, line 5: Here also.

Prabhaker24 (changes requested):
Can you try to generate deepcopy.go as well and add the corresponding ut's. Deepcopy might change as we are adding new probe struct in zookeepercluster_types.go

charts/zookeeper/values.yaml
anishakj: Reverted

anishakj:
> Can you try to generate deepcopy.go as well and add the corresponding ut's. Deepcopy might change as we are adding new probe struct in zookeepercluster_types.go

Added those changes

Prabhaker24 (approved):
LGTM

PR #293, opened 2021-01-20
Issue 291: Fixing post-install-upgrade hook

### LINKED ISSUE #291: ZK cluster post-install-upgrade hook does not wait for all replicas
### Description

The post-install-upgrade hook in zookeeper cluster chart does verify that all desired replicas have been deployed.
Since it checks for `status.replicas == status.readyReplicas`, the hook might pass, for example, when only two of three desired replicas have been deployed successfully, before the third replica even started creating.

### Importance

should-have

### Location

https://github.com/pravega/zookeeper-operator/blob/v0.2.9/charts/zookeeper/templates/post-install-upgrade-hooks.yaml#L66

The hook should rather use `spec.replicas` in the check instead of `status.replicas`, so that the number of ready replicas is compared to the desired number of replicas.

SrishT:

### Change log description
The post-install-upgrade hook in zookeeper cluster charts currently does not verify that all desired replicas have been deployed.

### Purpose of the change
Fixes #291 

### What the code does
The hook compares that the number of ready replicas is equal to the replica count mentioned within the spec, i.e. it checks that `.status.readyReplicas == .spec.replicas`

### How to verify it
The zookeeper cluster deployment should succeed only when all the desired replicas have come to ready state.

### DIFF
--- a/charts/zookeeper/templates/post-install-upgrade-hooks.yaml
+++ b/charts/zookeeper/templates/post-install-upgrade-hooks.yaml
@@ -63,7 +63,7 @@ data:
     set -e
     sleep 30
 
-    replicas=`kubectl get zk -n {{ .Release.Namespace }} {{ template "zookeeper.fullname" . }} -o jsonpath='{.status.replicas}'`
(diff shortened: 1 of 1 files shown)


anishakj (approved):
LGTM

PR #295, opened 2021-01-21
Issue 244: pod securityContext set in the zookeeper CRD, not taking effect

### LINKED ISSUE #244: pod securityContext set in the zookeeper CRD, not taking effect
### Description

Tried to configure securityContext in the CRD as below and it is not taking effect
spec:
  pod: 
    securityContext:
      runAsGroup: 1001
      runAsNonRoot: true
      runAsUser: 1001

### Importance

want to have securityContext at pod level to run as nonroot user

### Location

anishakj:

### Change log description

Added support for init containers, if security context is specified in pod spec, will give permission to that userid in the init containers

### Purpose of the change

 Fixes #244

### What the code does

Added support configuring securitycontext in pod spec. If non root user is specified, provide owner permissions to that user.

### How to verify it

Verified by providing security context in pod spec and without  the security context


charts/zookeeper/templates/zookeeper.yaml
```diff
@@ -7,6 +7,27 @@ metadata:
   labels:
 {{ include "zookeeper.commonLabels" . | indent 4 }}
 spec:
+  initContainers:
```
sarlaccpit, line 10: I think the logic that is present in the command of this init container is quite important, and should not be left to the submitter of the Zookeekeeper Cluster CR.

That init container logic is something that should be provided to anybody who submits a ZookeeperCluster for provisioning.  

If you look here: https://github.com/bitnami/charts/blob/master/bitnami/postgresql/templates/statefulset.yaml#L110, that initcontainer with similar logic is being embedded right into the statefulset.

In the postgresql case, it's part of the chart itself because the postgres instance is not a CR.   In here, I think it should be part of the zookeeper operator code that creates this statefulset  (generators.go, makeZkPodSpec) for the Zkcluster.    Meaning the chown logic etc. shouldn't be up to the end user.

There are additional nuances to this with OPenshift:   see how you have runAsUser: 0 in the securityContext of the init container ?    This could cause problems in some openshift environment if the serviceaccount for the pod (which applies to all containers of the pod including the init containers) has not been given access to the anyuid SCC.   (it is just not a guarantee at all).

I think a good path forward would be to imitate what postgresql did:

1. first, there have a boolean in to indicate where or not you want this volume permission initcontainer to be injected in the pod or not.
2. second, if you do want to have the initcontainer that sets permission, be aware that this might be in an openshift env., and so you might want to let Openshift decide what the "RunUserAs" is.   This comment explains it well https://github.com/bitnami/charts/blob/master/bitnami/postgresql/values.yaml#L71
(in their case, they use this "auto" keyword to indicate to leave the RunAsUser blank when they copy the security context object)

So either have a RunUserAs SC that gets applied to both the initcontainer and the main container.   Or a securityContext with NO RunAsUser, in both, to let Openshift provide you with a uid, that you then just chown things to.

And again, because that logic is tricky, best to have the zk operator provide this when it's building the statefulset than leave it open for the submitted of the zkcluster CR who will for sure get it wrong. :)

sarlaccpit (changes requested):
I think the initcontainer logic needs to be embedded in the zkoperator code itself and not left up to the user to provide/copy paste.

charts/zookeeper/templates/zookeeper.yaml
anishakj: I have changed the approach as you suggested. Please take a look

rm75 (approved):
LGTM. 
It should be documented that for Openshift, it is recommended to set the runAsUser value to auto.

Prabhaker24 (approved):
LGTM

pkg/zk/generators.go
```diff
@@ -175,13 +174,43 @@ func makeZkPodSpec(z *v1beta1.ZookeeperCluster, volumes []v1.Volume) v1.PodSpec
...
-	if reflect.DeepEqual(v1.PodSecurityContext{}, z.Spec.Pod.SecurityContext) {
+	if z.Spec.Pod.SecurityContext != nil {
 		podSpec.SecurityContext = z.Spec.Pod.SecurityContext
 	}
 	podSpec.NodeSelector = z.Spec.Pod.NodeSelector
 	podSpec.Tolerations = z.Spec.Pod.Tolerations
 	podSpec.TerminationGracePeriodSeconds = &z.Spec.Pod.TerminationGracePeriodSeconds
 	podSpec.ServiceAccountName = z.Spec.Pod.ServiceAccountName
+	if z.Spec.VolumePermissions == true && z.Spec.Pod.SecurityContext != nil {
+		runAsUser := int64(0)
```
sarlaccpit, line 185: I think that both the RunAsUser field of the init container and the main container both need to be some specific value from the values file (such as "0" but not necessarily), or auto (meaning randomized by openshift).

Hardcoding the runAsUser field of the init container to 0 is probably not what we want ultimately.  Because that means, if the init container runs as root, that essentially elevates the need for the whole pod (both containers) to run as root, since for example in Openshift, you'd need to give permission to the service account that runs the pod (zookeeper) access to the anyuid SCC.

Looking at the PostgreSQL chart, it shows [here ](https://github.com/bitnami/charts/blob/master/bitnami/postgresql/templates/statefulset.yaml#L112-L114) how it assigns a fully specified security context with RunAsUser if specified, or with the RunAsUser omitted if set to auto.

That is for the initcontainer itself.    Additionally, the target chown user that is used by the script to change the volume permission, is also following the same logic.  

https://github.com/bitnami/charts/blob/master/bitnami/postgresql/templates/statefulset.yaml#L86-L88

It does make things a bit confusing to have 2 security contexts (one for the initcontainer, and one for the main container); although it also brings flexbility.

For example, you could be in **scenario 1**:

Say your target user for the main container is 1001.
The the PV that is mounted inside the init container requires the user to be root in order to be able the permissions (usually due to some NFS server side configuration).  

In that case, you'd set the initcontainer's security context with RunAsUser: 0, and the maincontainer with security container: 1001.

The initcontainer script runs as root, ans chowns the volume to 1001.

**Scenario 2:**

You want your target to be running as 1001, and PV's permissions don't require the init container to run as root.  You can then set both to a security context with the 1001 context.    The initcontainer script runs as 1001 and chowns the volume to 1001.

**Scenario 3:**

You don't care what the runasuser is, you're running on Openshift and the volume doesn't require you to be root in order to change the permissions.

In that case; you would set the RunAsUser to auto; for both init and main container.  And let Openshift compute a uid when deploying the pod.   

The init container runs the script as user 123456, and chowns the target mount point to 123456

anishakj: @sarlaccpit  Addressed the comments. Please have a look

pkg/zk/generators.go
```diff
@@ -175,13 +174,54 @@ func makeZkPodSpec(z *v1beta1.ZookeeperCluster, volumes []v1.Volume) v1.PodSpec
...
+		if z.Spec.InitContainers == nil || z.Spec.InitContainers[0].SecurityContext == nil {
+			securityContext = &v1.SecurityContext{}
+			initContainerName = "initcontainer"
+		} else {
+			initContainerName = z.Spec.InitContainers[0].Name
+			securityContext = z.Spec.InitContainers[0].SecurityContext.DeepCopy()
+
+		}
+		var arg string
+		if podSpec.SecurityContext == nil || podSpec.SecurityContext.RunAsUser == nil {
```
sarlaccpit, line 197: I don't think `podSpec.SecurityContext` could be nil at this point.  `podSpec.SecurityContext.RunAsUser` could though.

charts/zookeeper/values.yaml
```diff
@@ -79,3 +83,7 @@ hooks:
 
 containers: []
 volumes: []
+#initContainers:
```
sarlaccpit, line 86: Since the initContainer is really specialized for the volume permission effort (in the code we actually look for initContainers[0] explicitly, I think it would be clearer to bind then together here and have a section like:
```
volumePermissions:
  enabled: true
  initContainer:
    securityContext:
      runasUser: 0  # and lose the auto keyword.  We probably just need a value or no value (which means auto)
```

charts/zookeeper/values.yaml
```diff
@@ -29,7 +30,10 @@ pod:
   # tolerations: []
   # env: []
   # annotations: {}
-  # securityContext: {}
+  # securityContext:
+    # Mention runAsUser as auto in case of openshift
```
sarlaccpit, line 34: I would lose the auto keyword (sorry I suggested that), and keep it simple:  Either specify a value for runAsUser, which can be 0 or any number, or don't specify a value, in which case it will be determined automatically.  (which is valid in non openshift env. also.  But is especially useful in Openshift envs)

pkg/zk/generators.go
anishakj: Done

charts/zookeeper/values.yaml
anishakj: Modified accordingly

charts/zookeeper/values.yaml
anishakj: Removed `auto`

charts/zookeeper/values.yaml
```diff
@@ -29,7 +29,10 @@ pod:
   # tolerations: []
   # env: []
   # annotations: {}
-  # securityContext: {}
+  # securityContext:
+    #runAsUser: 1000
+    #runAsGroup: 1000
+    #runAsNonRoot: false
```
sarlaccpit, line 35: I don't think `runAsNonRoot` is used anywhere.  We should probably remove the comment here to avoid confusion.

charts/zookeeper/values.yaml
```diff
@@ -79,3 +82,13 @@ hooks:
...
 containers: []
 volumes: []
+
+## Init containers parameters:
+## volumePermissions: Change the owner of the persistent volume mountpoint to RunAsUser:RunAsGroup
+
+volumePermissions:
+  enabled: false
+  #initContainer:
+    #name: initcontainer
```
sarlaccpit, line 92: I wonder if giving an initContainer.name field here adds anything.   Since that specific init container is actually fully managed by the operator, maybe hardcode/set a name for it in the code, no matter what, like you are doing at line 190 of generators.go. but maybe something more meaningful like volume-perms-init or similar.

pkg/zk/generators.go
```diff
@@ -175,13 +174,54 @@ func makeZkPodSpec(z *v1beta1.ZookeeperCluster, volumes []v1.Volume) v1.PodSpec
...
-	if reflect.DeepEqual(v1.PodSecurityContext{}, z.Spec.Pod.SecurityContext) {
+	if z.Spec.Pod.SecurityContext != nil {
 		podSpec.SecurityContext = z.Spec.Pod.SecurityContext
 	}
 	podSpec.NodeSelector = z.Spec.Pod.NodeSelector
 	podSpec.Tolerations = z.Spec.Pod.Tolerations
 	podSpec.TerminationGracePeriodSeconds = &z.Spec.Pod.TerminationGracePeriodSeconds
 	podSpec.ServiceAccountName = z.Spec.Pod.ServiceAccountName
+
+	if z.Spec.VolumePermissions == true && z.Spec.Pod.SecurityContext != nil {
```
sarlaccpit, line 185: It sounds like a nit but I think that we should insert a warning after line 224, if VolumePermissions was set to true but no z.Spec.Pod.SecurityContext was supplied, it sounds like the user intended to use the new init container, but forgot to set the z.Spec.Pod.SecurityContext and so at the end of the day didn't get the full intended effect.

Logging it with a warning would bring attention to what is probably a misconfiguration of the Zookeeper CR they dropped in.

charts/zookeeper/values.yaml
anishakj: Changed

charts/zookeeper/values.yaml
anishakj: Done

pkg/zk/generators.go
anishakj: Added log message

pkg/zk/generators_test.go
```diff
@@ -222,6 +223,89 @@ var _ = Describe("Generators Spec", func() {
...
+			BeforeEach(func() {
+				z := &v1beta1.ZookeeperCluster{
+					ObjectMeta: metav1.ObjectMeta{
+						Name:      "example",
+						Namespace: "default",
+					},
+				}
+				z.WithDefaults()
+				z.Spec.VolumePermissions = true
+				runAsNonRoot := false
```
sarlaccpit, line 268: runAsRoot has been removed so either the test needs to be updated or removed if it's no long relevant.

charts/zookeeper/templates/zookeeper.yaml
```diff
@@ -41,6 +41,15 @@ spec:
   {{- if .Values.volumes }}
   volumes:
 {{ toYaml .Values.volumes | indent 4 }}
+  {{- end }}
+  volumePermissions: {{ .Values.volumePermissions.enabled }}
+  {{- if .Values.volumePermissions.enabled }}
+  initContainers:
+  - name: volume-perms-init
```
sarlaccpit, line 48: the name field is no longer needed since it's being set to a fixed value in the code.

pkg/zk/generators_test.go
anishakj: this unit test is added to cover the case runasuser is nil, but security context is not nil

charts/zookeeper/templates/zookeeper.yaml
anishakj: we cant exclude name field  in order to comply with openAPi schema validation. Container name is a mandatory field

anishakj:
Closing this PR as different approach has to be taken to fix the issue

PR #301, opened 2021-02-04
Issue 300: Zookeeper Operator pre-delete hook

### LINKED ISSUE #300: No pre-delete hook in zookeeper-operator chart
### Description

The zookeeper-operator helm chart does not have a pre-delete hook in it. 
As a result, the operator can be uninstalled with orphan custom resources hanging around.

The orphan custom resources cannot be deleted, because they have finalizers that are supposed to be cleared by the operator. However, the operator is gone at this point. 

Even the operator is reinstalled, it most probably won't be able to clean up the above custom resource if its parent namespace had been marked for deletion and is stuck in the terminated state because of that orphan custom resource.

### Importance

should-have

### Location

zookeeper-operator helm chart

Please, add a pre-delete hook to the operator chart  that would verify that all custom resources have been cleaned up before deleting the operator.

SrishT:

### Change log description
Provides a pre-delete hook which ensures that custom resources are deleted before deleting the operator.

### Purpose of the change
Fixes #300 

### What the code does
Allows the operator to be deleted via helm only if the zookeepercluster custom resource has already been deleted, not otherwise.

### How to verify it
- If you try to delete the zookeeper operator before having deleted the zookeeper cluster resource, the pre-delete job will fail, thereby not allowing the operator from being deleted.
- If you try to delete the zookeeper operator after the zookeeper cluster resource has been cleaned up successfully, the operator will also be successfully deleted.

### DIFF
--- a/charts/zookeeper-operator/templates/pre-delete-hooks.yaml
+++ b/charts/zookeeper-operator/templates/pre-delete-hooks.yaml
@@ -1,0 +1,110 @@
+kind: ClusterRole
+apiVersion: rbac.authorization.k8s.io/v1beta1
+metadata:
+  name: {{ template "zookeeper-operator.fullname" . }}-pre-delete
+  annotations:
+    "helm.sh/hook": pre-delete
+    "helm.sh/hook-weight": "1"
+    "helm.sh/hook-delete-policy": hook-succeeded, before-hook-creation, hook-failed
+rules:
+- apiGroups:
+  - zookeeper.pravega.io
+  resources:
+  - "*"
+  verbs:
+  - get
+  - list
+---
+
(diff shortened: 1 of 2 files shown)


jkhalack (approved):
LGTM

anishakj (changes requested):
Do we need to ensure role and role bindings deleted in addition to the cluster role?

SrishT:
> Do we need to ensure role and role bindings deleted in addition to the cluster role?

@anishakj we are only ensuring that the zookeepercluster custom resource has only been deleted before triggering the delete. We are not checking for the RBAC being already deleted since it is going to be deleted by this chart itself.

anishakj (approved):
LGTM

PR #305, opened 2021-02-25
Issue 299: Fix incorrect paths in config spec

### LINKED ISSUE #299: Helm config spec cannot be set because the some paths are incorrect
### Description

In the zookeeper helm chart, some config specs cannot be set because the paths are incorrect.
They should be;

```
    {{- if .Values.config.xxxx}}
    xxxx: {{ .Values.config.xxxx }}
    {{- end }}
```

But in some lines, they are;

```
    {{- if .Values.config.xxxx}}
    xxxx: {{ .Values.xxxx }}
    {{- end }}
```

### Location

https://github.com/pravega/zookeeper-operator/blob/b50eaa4565b8643cecf9fb1216a5470d8efb4423/charts/zookeeper/templates/zookeeper.yaml#L91-L137

It is necessary to set correct paths.

alim-akbashev:
### Change log description

Fix Helm config spec cannot be set because the some paths are incorrect

### Purpose of the change

Fixes #299 

### What the code does

Typos fixed

### How to verify it

Changes in snapSizeLimitInKb, maxCnxns, maxClientCnxns, minSessionTimeout, maxSessionTimeout, autoPurgeSnapRetainCount should be applied to configMap and to zoo.cfg

### DIFF
--- a/charts/zookeeper/templates/zookeeper.yaml
+++ b/charts/zookeeper/templates/zookeeper.yaml
@@ -111,22 +111,22 @@ spec:
     commitLogCount: {{ .Values.config.commitLogCount }}
     {{- end }}
     {{- if .Values.config.snapSizeLimitInKb }}
-    snapSizeLimitInKb: {{ .Values.snapSizeLimitInKb }}
+    snapSizeLimitInKb: {{ .Values.config.snapSizeLimitInKb }}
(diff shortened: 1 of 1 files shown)


anishakj (approved):
LGTM

PR #306, opened 2021-02-26
Alternate zk versions

### LINKED ISSUE #286: Older Zookeeper versions
### Description

Question: Is it possible to generate/deploy older versions of zookeeper with this operator (ie 3.4.x)? Or there are dependencies related with Zookeeper? Thanks!

cognifloyd:
### Change log description

Added a simple way to use a different version of ZooKeeper.

### Purpose of the change

Fixes #286

### What the code does

By running `make zk-build-image ZOOKEEPER_VERSION=3.5.5` you could, for example, build a zookeeper image with version 3.5.5 instead of the default 3.6.1. 

Note, however, that 3.5.5 is the oldest version that this Operator can support. It depends on the dynamic reconfig feature which was added in 3.5.3. As 3.5.5 is the first stable release of the 3.5.y series, so it is the first release to include this feature.

### How to verify it

Build it. Use it.
You'll need to pass in the name/version of your custom docker image when configuring the operator, eg via values in the Helm chart.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -11,6 +11,8 @@ SHELL=/bin/bash -o pipefail
 PROJECT_NAME=zookeeper-operator
 EXPORTER_NAME=zookeeper-exporter
 APP_NAME=zookeeper
+# also update ZOOKEEPER_VERSION in docker/Dockerfile
+ZOOKEEPER_VERSION ?= 3.6.1
 KUBE_VERSION=1.17.5
 REPO=pravega/$(PROJECT_NAME)
 TEST_REPO=testzkop/$(PROJECT_NAME)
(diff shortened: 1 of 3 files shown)


cognifloyd:
I built this trying to use ZooKeeper 3.4.11 with the Operator. As that is not possible, I will be using something else. Please feel free to edit this PR and build off of it.

lunarfs:
@cognifloyd I will try to take this further, I have been working on https://github.com/lunarfs/zookeeper-operator/tree/zookeeper_dynamic_version as I need 3.5.8 and that should indeed be possible. So i have build  an image that i will be testing.. once that is completed i will take a look at making a PR or taking over this .. i have no plan yet.. but it will happen.

PR #308, opened 2021-03-01
Issue 307: Provide a way for customization of annotations for service…

### LINKED ISSUE #307: Custom annotations for services created by operator, optional LoadBalancer
### Description

It would be great to have ability to attach custom annotations for services created by the Operator, and to make attaching LoadBalancer for AdminServer service optional

### Importance

This is especially important  for admin-server service, since by default LoadBalancer with default configuration is being attached. For my environment it conflicts with automated provisioning of resources which requires to have specific annotation on resources created by 3rd-party tools. So it's kind of blocker for me.
Additionally, I believe it would be great to make attaching LoadBalancer optional since LB costs money (at least in AWS EKS) and some people do not really use it outside of k8s.

### Location

pkg/apis/zookeeper/v1beta1/zookeepercluster_types.go
pkg/zk/generators.go

I'm going to create PR if nobody minds

alim-akbashev:
### Change log description

A way for customization of annotations for services and for making LoadBalancer optional for AdminServer service has been added.

### Purpose of the change

Fixes #307 

### What the code does

- clientService, headlessService, adminServerService were added to CRD
- support of new objects added to generators.go
- new code covered with tests

### How to verify it

Adding annotations to clientService/headlessService/adminServerService should lead to attaching corresponding annotations to k8s Service's.
Adding `adminServerService.internal: true` should lead to creating k8s Service with ClusterIP without LoadBalancer.

### DIFF
--- a/charts/zookeeper-operator/templates/zookeeper.pravega.io_zookeeperclusters_crd.yaml
+++ b/charts/zookeeper-operator/templates/zookeeper.pravega.io_zookeeperclusters_crd.yaml
@@ -2426,9 +2426,9 @@ spec:
               description: AdminServerService defines the policy to create AdminServer Service
                 for the zookeeper cluster.
               properties:
-                internal:
-                  description: Internal specifies if LoadBalancer should be created for
-                    the AdminServer. False means LoadBalancer will be created, true means ClusterIP
+                external:
+                  description: External specifies if LoadBalancer should be created for
+                    the AdminServer. True means LoadBalancer will be created, false means ClusterIP
                     will be used. Default is false.
                   type: boolean
                 annotations:
--- a/charts/zookeeper-operator/templates/zookeeper.pravega.io_zookeeperclusters_crd.yaml
+++ b/charts/zookeeper-operator/templates/zookeeper.pravega.io_zookeeperclusters_crd.yaml
@@ -2400,6 +2400,44 @@ spec:
                     type: object
                   type: array
               type: object
+            clientService:
+              description: ClientService defines the policy to create client Service
+                for the zookeeper cluster.
+              properties:
+                annotations:
+                  additionalProperties:
+                    type: string
+                  description: Annotations specifies the annotations to attach to
+                    client Service the operator creates.
+                  type: object
+              type: object
+            headlessService:
+              description: HeadlessService defines the policy to create headless Service
+                for the zookeeper cluster.
+              properties:
+                annotations:
+                  additionalProperties:
+                    type: string
+                  description: Annotations specifies the annotations to attach to
+                    headless Service the operator creates.
+                  type: object
+              type: object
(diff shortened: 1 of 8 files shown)


SrishT:
Thank you for this PR @alim-akbashev. We will review it shortly.

SrishT (changes requested):
The changes look good. The only problem is that the parameters within the charts have been added in a way that it fails with the following error if the annotations are left empty
```
Error: template: zookeeper/templates/zookeeper.yaml:91:18: executing "zookeeper/templates/zookeeper.yaml" at <.Values.clientService.annotations>: nil pointer evaluating interface {}.annotations
```
Please make changes such that it accounts for these fields being empty as well.

alim-akbashev:
thank you for review @SrishT! please advise me if could apply better fix.

SrishT:
> thank you for review @SrishT! please advise me if could apply better fix.

@alim-akbashev it might help to have an if check around the [clientService](https://github.com/pravega/zookeeper-operator/blob/ffc2b9fb04911af7b22dc394057424cfa1c55608/charts/zookeeper/templates/zookeeper.yaml#L90), [headlessService](https://github.com/pravega/zookeeper-operator/blob/ffc2b9fb04911af7b22dc394057424cfa1c55608/charts/zookeeper/templates/zookeeper.yaml#L95) and [adminServerService](https://github.com/pravega/zookeeper-operator/blob/ffc2b9fb04911af7b22dc394057424cfa1c55608/charts/zookeeper/templates/zookeeper.yaml#L100), like we have done in case of all other fields, so that that particular portion is populated only in case some value is specified for those fields by the user.

alim-akbashev:
makes sense, thanks!

SrishT (changes requested):
Instead of exposing the AdminServer Service as an external loadbalancer service by default, it might be more desirable to expose it as an internal service. 
Also it might be better to specify the field in the following manner
```
adminServerService:
    external: false
    annotations: {}
```
where the default value of the field `adminServerService.external` would be false, so that even when the user does not provide any value for this field, it is exposed as an internal service by default.

alim-akbashev:
Definitely makes sense, i've thought about that as well, just tried to support current behaviour.
Anyway, will make it done on Monday, thanks!

alim-akbashev:
Done. `adminServerService.internal` has been replaced with `adminServerService.external` (false by default).
Unfortunately CI fails sometimes. It seems like that does not relate to changes in the PR since it fails on master branch as well (and has been failing few times on my branch, re-run helped that time):
```
image_pullsecret_test.go:56: 
        Unexpected error:
            <*errors.errorString | 0xc000113880>: {
                s: "timed out waiting for the condition",
            }
            timed out waiting for the condition
        occurred
```

SrishT (approved):
LGTM

Issue #312, opened 2021-03-15
[BUG] PVC can be mistakenly deleted when reading from stale apiserver


srteam2020:
### Description

We find that zookeeper-operator could accidentally delete the PVC of the zookeeper pod when zookeeper-operator experiences a restart and talks to one stale apiserver in an HA k8s cluster. After several rounds of inspection, we find the root cause is the staleness of the events fed from the apiserver to the controller.

More concretely, the controller performs reconcile() according to the events watched from apiserver, and the watch is served by the apiserver’s local cache without any strong consistency guarantee. If that apiserver's watch cache is stale (due to some network partition issue), the controller will perform reconcile() according to the stale events from the apiserver, which can lead to failures.

Particularly, if the controller receives a stale update event indicating "ZookeeperCluster has non-zero deletion timestamp" from the stale apiserver, it will believe a deletion is happening and decides to delete the PVC of that ZookeeperCluster. Note that such a stale event actually comes from the deletion of a previous ZookeeperCluster that has a different UID than the currently running one. However, the controller lists all related PVC only using the name and namespace:
```
func (r *ReconcileZookeeperCluster) getPVCList(instance *zookeeperv1beta1.ZookeeperCluster) (pvList corev1.PersistentVolumeClaimList, err error) {
	selector, err := metav1.LabelSelectorAsSelector(&metav1.LabelSelector{
		MatchLabels: map[string]string{"app": instance.GetName()},
	})
	pvclistOps := &client.ListOptions{
		Namespace:     instance.Namespace,
		LabelSelector: selector,
	}
	pvcList := &corev1.PersistentVolumeClaimList{}
	err = r.client.List(context.TODO(), pvcList, pvclistOps)
	return *pvcList, err
}
```

If the currently running ZookeeperCluster uses the same name as the previously deleted one (e.g., we delete and recreate a ZookeeperCluster), the controller will mistakenly think the listed PVCs belong to the "deleted ZookeeperCluster", and invokes `deletePVC` to delete the PVC.

We list concrete reproduction steps as below:
1. Run the controller in a HA k8s cluster and create a ZookeeperCluster zkc. The controller is talking to apiserver1 (which is not stale).
2. Delete zkc. Apiserver1 will send the update events with a non-zero deletion timestamp to the controller and the controller will delete the related PVC of zkc during reconcile. Meanwhile, apiserver2 is partitioned so its watch cache stops at the moment that zkc is tagged with a deletion timestamp.
[...]

PR #313, opened 2021-03-20
Issue 312: list PVC with ZookeeperCluster uid

srteam2020:
### Change log description

Add UID as a label for each PVC in `MakeStatefulSet` and list PVC using the UID in `getPVCList`.

### Purpose of the change

Fix https://github.com/pravega/zookeeper-operator/issues/312.

### What the code does

- `MakeStatefulSet`: One more label, uid, when creating VolumeClaimTemplates
- `getPVCList`: List the PVC with the uid of the current ZookeeperCluster

### How to verify it

- Normal PVC creation/deletion without node failures can still work
- As described in https://github.com/pravega/zookeeper-operator/issues/312, the reproduction steps will not trigger PVC deletion


anishakj (approved):
LGTM

PR #317, opened 2021-04-09
Remove the skipACL property from zoo.cfg; users can add via the SERVER_JVMFLAGS env var if needed

### LINKED ISSUE #316: skipACL is hardcoded to "yes" and cannot be overridden via env vars
### Description

The `skipACL` property is set to yes in `zoo.cfg`, see:
https://github.com/pravega/zookeeper-operator/blob/b1519aec85c52bd946eaa91b4bcf58e1767210f2/pkg/zk/generators.go#L250

This prevents consumer applications (Solr in my case) from being able to guard sensitive znodes. Does the operator rely on ACLs not being enforced? 

I tried overriding the value from `zoo.cfg` with an env var:
```
        env:
          - name: SERVER_JVMFLAGS
            value: -Dzookeeper.skipACL=no
```
This should work because of:
https://github.com/apache/zookeeper/blob/release-3.6.1/zookeeper-server/src/main/java/org/apache/zookeeper/server/ZooKeeperServer.java#L130

And the system prop is getting passed to the Java process correctly:
```
... -Dzookeeper.skipACL=no -Dcom.sun.management.jmxremote -Dcom.sun.management.jmxremote.local.only=false org.apache.zookeeper.server.quorum.QuorumPeerMain /data/conf/zoo.cfg
```

However, this doesn't work b/c `QuorumPeerConfig#parseProperties` runs before `ZooKeeperServer`'s static initializer block gets called. Unfortunately, parseProperties does this for any additional properties in the zoo.cfg:
```
            } else {
                System.setProperty("zookeeper." + key, value);
            }
```
See: https://github.com/apache/zookeeper/blob/release-3.6.1/zookeeper-server/src/main/java/org/apache/zookeeper/server/quorum/QuorumPeerConfig.java#L408

### Importance

must-have! This prevents enforcing ACLs established by our application.

### Location

https://github.com/pravega/zookeeper-operator/blob/b1519aec85c52bd946eaa91b4bcf58e1767210f2/pkg/zk/generators.go#L250

The operator should allow this config property to be changed.

thelabdude:
### Change log description

Don't force `skipACL` to `yes` in zoo.cfg; if users need this, they can set via the SERVER_JVMFLAGS env var if needed, e.g. `SERVER_JVMFLAGS=-Dzookeeper.skipACL=yes`

The operator should not be disabling security features without good reason.

### Purpose of the change

Fixes #316 

### What the code does

Removes the `skipACL=yes` from the generated zoo.cfg, which disables ACL enforcement and is a security risk.

Removing this property breaks multi-node ensembles, but seems like the operator should just read creds from a K8s secret vs. disabling the entire ACL / security system?

### How to verify it

TODO ...

### DIFF
--- a/pkg/zk/generators.go
+++ b/pkg/zk/generators.go
@@ -226,7 +226,6 @@ func makeZkConfigString(s v1beta1.ZookeeperClusterSpec) string {
 		"dataDir=/data\n" +
 		"standaloneEnabled=false\n" +
 		"reconfigEnabled=true\n" +
-		"skipACL=yes\n" +
 		"metricsProvider.className=org.apache.zookeeper.metrics.prometheus.PrometheusMetricsProvider\n" +
 		"metricsProvider.httpPort=7000\n" +
 		"metricsProvider.exportJvmInfo=true\n" +
--- a/pkg/zk/generators_test.go
+++ b/pkg/zk/generators_test.go
@@ -79,10 +79,6 @@ var _ = Describe("Generators Spec", func() {
 					Ω(cfg).To(ContainSubstring("reconfigEnabled=true\n"))
 				})
 
-				It("should set skipACL to 'yes'", func() {
-					Ω(cfg).To(ContainSubstring("skipACL=yes"))
-				})
-
 				It("should set initLimit to '10'", func() {
 					Ω(cfg).To(ContainSubstring("initLimit=10\n"))
 				})


thelabdude:
heh ... spoke too soon, I only tried a single node ... adding more nodes, now my 2nd pod goes into crash backoff:
```
+ java -Dlog4j.configuration=file:/conf/log4j-quiet.properties -jar /root/zu.jar add dev-solrcloud-zookeeper-client:2181 2 'dev-solrcloud-zookeeper-1.dev-solrcloud-zookeeper-headless.dev.svc.cluster.local:2888:3888:observer;2181' /data/zoo.cfg.dynamic
Connecting to Zookeeper dev-solrcloud-zookeeper-client:2181
Error performing zookeeper reconfiguration:
org.apache.zookeeper.KeeperException$NoAuthException: KeeperErrorCode = NoAuth
	at org.apache.zookeeper.KeeperException.create(KeeperException.java:120)
	at org.apache.zookeeper.KeeperException.create(KeeperException.java:54)
	at org.apache.zookeeper.admin.ZooKeeperAdmin.reconfigure(ZooKeeperAdmin.java:246)
	at io.pravega.zookeeper.MainKt.reconfigure(Main.kt:122)
	at io.pravega.zookeeper.MainKt.runAdd(Main.kt:145)
	at io.pravega.zookeeper.MainKt.main(Main.kt:30)
```

thelabdude:
Not going to work on this as it seems the operator relies on skipACL

PR #326, opened 2021-05-10
fix issue-314 by checking zookeeperCluster resource version before updating sts

### LINKED ISSUE #314: [BUG] Reading stale ZookeeperCluster spec/status can lead to undesired pod and PVC deletion
### Description

We find that zookeeper-operator could accidentally delete zookeeper pod and its PVC when the controller reads stale information about ZookeeperCluster.

More concretely, we observe that if a ZookeeperCluster goes through scale down and then scale up, and one of the apisevers gets slow (or partitioned from others), the (restarted) controller may read the stale status/spec information of the ZookeeperCluster. The stale information can trigger the PVC deletion in`reconcileFinalizers` and statefulset scale down (zookeeper pod deletion) in `reconcileStatefulSet`.

We list concrete reproduction steps as below:
1. Run the controller in a HA k8s cluster and create a ZookeeperCluster `zkc` with `replicas=2`. The controller is talking to apiserver1 (which is not stale). There will be two zookeeper pods `zk1` and `zk2` and two PVC in the cluster.
2. Scale `zkc` down (by setting `replicas=1`). After getting stable, there will only be one zookeeper pod `zk1` and one PVC.  Meanwhile, apiserver2 gets partitioned so its watch cache stops at the moment that `zkc` has `replicas=1`.
3. Scale `zkc` up (by setting `replicas=2`). Now a new `zk2` and its PVC get back.
4. After experiencing a crash, the restarted controller talks to the stale apiserver2. From apiserver2's watch cache, the controller finds that `zkc`'s `Status.ReadyReplicas` and `Spec.Replicas` are both 1, which is lower than the PVC count (i.e., 2). Inside `cleanupOrphanPVCs` the controller will treat the PVC of `zk2` as an orphan PVC and delete it. Later in `Updating StatefulSet`, the controller will also set the `Replicas` of the statefulset back to 1, which will trigger deletion of `zk2`.

### Importance

blocker: The unexpected PVC and pod deletion caused by reading stale data from the apiserver can further lead to data loss or availability issues.

### Location

zookeepercluster_controller.go

We are willing to help alleviate this problem by issuing a PR.
It is hard to fully avoid this issue because the controller has no clue whether it is reading the stale information of the ZookeeperCluster or not, given the controller is supposed to be stateless. However, some sanity check can help prevent this issue in certain cases and ensure the controller is not going to perform undesired deletion before the updated information is populated to the controller:
1. Before issuing PVC deletion in `cleanupOrphanPVCs`, we can first check the number of zookeeper pods and see whether the number is the same as `Spec.Replicas` of the ZookeeperCluster. If not, we know at least one of them is wrong and the PVC may not be an orphan. The sanity check is helpful if the pod information is fresh (especially when the pod information is retrieved from a different apiserver/etcd than the  ZookeeperCluster information).
2. Before updating the statefulset in `reconcileStatefulSet`, we may want to check `ZookeeperCluster.Spec.Replicas`, `ZookeeperCluster.Status.Replicas` and `StatefulSet.Spec.Replicas`. For example, it is a sign of abnormality if `ZookeeperCluster.Spec.Replicas != StatefulSet.Spec.Replicas` but  `ZookeeperCluster.Spec.Replicas == ZookeeperCluster.Status.Replicas`. Logging a warning and returning from reconcile() with error in such case can be helpful.

We are willing to issue a PR to add the above checks if they are not considered too expensive.

srteam2020:
### Change log description
The PR makes the following changes:
- label the current RV of the zookeeperCluster to the sts when creation
- check whether the current RV of the zookeeperCluster is larger than the RV labeled to the sts before updating the sts

### Purpose of the change

Fixes #314

### DIFF
--- a/pkg/controller/zookeepercluster/zookeepercluster_controller.go
+++ b/pkg/controller/zookeepercluster/zookeepercluster_controller.go
@@ -233,6 +233,10 @@ func (r *ReconcileZookeeperCluster) reconcileStatefulSet(instance *zookeeperv1be
 	} else if err != nil {
 		return err
 	} else {
+		// check whether zookeeperCluster is updated before updating the sts
+		if ! r.zookeeperClusterUpdated(instance, sts) {
+			return fmt.Errorf("Staleness: cr.ResourceVersion %s is smaller than labeledRV %s", instance.ResourceVersion, sts.Labels["owner-rv"])
+		}
 		foundSTSSize := *foundSts.Spec.Replicas
 		newSTSSize := *sts.Spec.Replicas
 		if newSTSSize != foundSTSSize {
@@ -253,10 +257,6 @@ func (r *ReconcileZookeeperCluster) reconcileStatefulSet(instance *zookeeperv1be
 			data := "CLUSTER_SIZE=" + strconv.Itoa(int(newSTSSize))
 			r.log.Info("Updating Cluster Size.", "New Data:", data, "Version", version)
 			r.zkClient.UpdateNode(path, data, version)
-		}
-		// check whether zookeeperCluster is updated before updating the sts
-		if ! r.zookeeperClusterUpdated(instance, sts) {
-			return fmt.Errorf("Staleness: cr.ResourceVersion %s is smaller than labeledRV %s", instance.ResourceVersion, sts.Labels["owner-rv"])
 		}
 		err = r.updateStatefulSet(instance, foundSts, sts)
 		if err != nil {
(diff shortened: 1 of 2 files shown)


anishakj:
@srteam2020 Thanks for raising PR, Could you please add unit tests to increase the test coverage for the new function added?

srteam2020:
@anishakj Thanks for the reply! I will add an unit test for the newly added function and push again later.

anishakj:
> @anishakj Thanks for the reply! I will add an unit test for the newly added function and push again later.

@srteam2020 Any updates on adding unit  tests?

srteam2020:
Hi @anishakj I was busy on some ddls these two weeks. I will work on it this weekend.

anishakj (approved):
LGTM

PR #327, opened 2021-05-14
Issue 328: Add support of customized volumeMounts for zookeeper container

### LINKED ISSUE #328: Support customized volumeMounts for zookeeper container
### Description

Need to support customized volumeMounts for zk container so that the customized volumes which user set can be mounted to the zk container.

### Importance

ECS ObjectScale need this feature so that zookeeper component can be intergrated with the rsyslog logging injection. 

### Location

Operator code & related templates

CraneShiEMC:
### Change log description

Added a new field volumeMounts in ZookeeperClusterSpec so that users can set customized volumeMounts for zookeeper container to mount customized volumes user specified.

### Purpose of the change

Fixes #328 

### What the code does

Added a new field volumeMounts in ZookeeperClusterSpec and the customized volumeMounts specified can be added to the zookeeper container.

### How to verify it

Verified that able to configure customized volumeMounts for zk container and successfully mounted the customized volumes to the zk container.


pkg/zk/generators.go
```diff
@@ -142,10 +142,10 @@ func makeZkPodSpec(z *v1beta1.ZookeeperCluster, volumes []v1.Volume) v1.PodSpec
 				Exec: &v1.ExecAction{Command: []string{"zookeeperLive.sh"}},
 			},
 		},
-		VolumeMounts: []v1.VolumeMount{
+		VolumeMounts: append(z.Spec.VolumeMounts, []v1.VolumeMount{
```
realAaronWu, line 145: Is this Spec.Volume Mountsinitialized by default?

realAaronWu:
Please add UT coverage for the Spec.VolumeMounts.

pkg/zk/generators.go
CraneShiEMC: Spec.VolumeMounts will be null by default and will not affect the original inherent volumeMounts of zk container. Only when we set values of spec.volumeMount in our charts, this will take effect and merge with original inherent volumemounts of zk container.

CraneShiEMC: > Is this Spec.VolumeMounts initialized by default?

Spec.VolumeMounts will be null by default and will not affect the original inherent volumeMounts of zk container. Only when we set values of spec.volumeMount in our charts, this will take effect and merge with original inherent volumemounts of zk container.

CraneShiEMC:
> Please add UT coverage for the Spec.VolumeMounts.

Added now

anishakj (approved):
LGTM

PR #330, opened 2021-05-19
Issue 329: CRD apiVersion check

### LINKED ISSUE #329: apiextensions.k8s.io/v1beta1 apiVersion removed on k8s v1.19.0
### Description

The apiextensions.k8s.io/v1beta1 apiVersion used in the CustomResourceDefinition was removed on k8s v1.19.0

### Importance

The issue is a blocker to install the chart properly on k8s clusters with version v1.19.0

### Location

https://github.com/pravega/zookeeper-operator/blob/master/charts/zookeeper-operator/templates/zookeeper.pravega.io_zookeeperclusters_crd.yaml#L2

I suggest the issue should be fixed by checking the apiVersion with .Capabilities.APIVersions.Has
```
{{- if .Capabilities.APIVersions.Has "apiextensions.k8s.io/v1" }}
apiVersion: apiextensions.k8s.io/v1
{{- else }}
apiVersion: apiextensions.k8s.io/v1beta1
{{- end }}
```

lspecian-olx:
### Change log description

- Add the capability to install the chart in eks versions that don't support apiVersion=apiextensions.k8s.io/v1beta1 anymore

### Purpose of the change
Closes #329 

### What the code does

The code is using .Capabilities.APIVersions.Has to check for the apiVersion apiextensions.k8s.io/v1, if it finds it it uses to create the CRD, if it doesn't it falls back to apiextensions.k8s.io/v1beta1

### How to verify it

You are able to install the chart with the CRD in eks v1.19.0 version from which apiextensions.k8s.io/v1beta1 was removed

### DIFF
--- a/charts/zookeeper-operator/templates/pre-delete-hooks.yaml
+++ b/charts/zookeeper-operator/templates/pre-delete-hooks.yaml
@@ -1,6 +1,6 @@
 {{- if .Values.hooks.delete }}
 kind: ClusterRole
-apiVersion: rbac.authorization.k8s.io/v1beta1
+apiVersion: rbac.authorization.k8s.io/v1
 metadata:
   name: {{ template "zookeeper-operator.fullname" . }}-pre-delete
   annotations:
@@ -18,7 +18,7 @@ rules:
 ---
 
 kind: ClusterRoleBinding
-apiVersion: rbac.authorization.k8s.io/v1beta1
+apiVersion: rbac.authorization.k8s.io/v1
 metadata:
   name: {{ template "zookeeper-operator.fullname" . }}-pre-delete
   annotations:
--- a/charts/zookeeper-operator/templates/zookeeper.pravega.io_zookeeperclusters_crd.yaml
+++ b/charts/zookeeper-operator/templates/zookeeper.pravega.io_zookeeperclusters_crd.yaml
@@ -1,5 +1,9 @@
 {{- if .Values.crd.create }}
+{{- if .Capabilities.APIVersions.Has "apiextensions.k8s.io/v1" }}
+apiVersion: apiextensions.k8s.io/v1
+{{- else }}
 apiVersion: apiextensions.k8s.io/v1beta1
+{{- end }}
 kind: CustomResourceDefinition
 metadata:
   name: zookeeperclusters.zookeeper.pravega.io


anishakj:
@lspecian-olx Thanks for the contribution. Do we need similar check here as well https://github.com/lspecian-olx/zookeeper-operator/blob/issue-329-crd-apiversion-check/charts/zookeeper-operator/templates/pre-delete-hooks.yaml#L3 ?

lspecian-olx:
> @lspecian-olx Thanks for the contribution. Do we need similar check here as well https://github.com/lspecian-olx/zookeeper-operator/blob/issue-329-crd-apiversion-check/charts/zookeeper-operator/templates/pre-delete-hooks.yaml#L3 ?

Indeed we need a similar check as well, I made a new commit including the change

charts/zookeeper-operator/templates/pre-delete-hooks.yaml
```diff
@@ -18,7 +22,11 @@ rules:
 ---
 
 kind: ClusterRoleBinding
+{{- if .Capabilities.APIVersions.Has "rbac.authorization.k8s.io/v1" }}
```
SrishT, line 25: The apiVersion for the `ClusterRole` and `ClusterRoleBinding`  can be changed to `rbac.authorization.k8s.io/v1` and we can remove the check from there, since we are anyways using that version in the other files.

anishakj:
@lspecian-olx  could you please address the review comment?

charts/zookeeper-operator/templates/pre-delete-hooks.yaml
lspecian-olx: I believe this would affect the compatibility with k8s versions 1.15.0 and below, perhaps the other files should also have the check?

anishakj: In our README, we have mentioned requirement as ` Kubernetes v1.15.0+`

lspecian-olx: well I guess for someone running k8s 1.15.0, this is the least of the problem, I'll change the commit removing the check

anishakj (approved):
LGTM

charts/zookeeper-operator/templates/pre-delete-hooks.yaml
lspecian-olx: If the requirement k8s v1.15.0+ we should probably check in the other files as well, I can change this, also I double checked, in fact the rbac.authorization.k8s.io/v1beta1 was only deprecated on 1.17.0, so I assume rbac.authorization.k8s.io/v1 is available only from 1.17.0+

the other option would be to update the requirements

anishakj: have  seen that `rbac.authorization.k8s.io/v1` is supported in cluster with version  `v1.15.10`

SrishT (approved):
LGTM

anishakj:
@lspecian-olx  when i tried to install Operator on a cluster with `v1.19.9` version, got the below error. Are you also seeing same?

```Error: unable to build kubernetes objects from release manifest: error validating "": error validating data: [ValidationError(CustomResourceDefinition.spec): unknown field "additionalPrinterColumns" in io.k8s.apiextensions-apiserver.pkg.apis.apiextensions.v1.CustomResourceDefinitionSpec, ValidationError(CustomResourceDefinition.spec): unknown field "subresources" in io.k8s.apiextensions-apiserver.pkg.apis.apiextensions.v1.CustomResourceDefinitionSpec, ValidationError(CustomResourceDefinition.spec): unknown field "validation" in io.k8s.apiextensions-apiserver.pkg.apis.apiextensions.v1.CustomResourceDefinitionSpec, ValidationError(CustomResourceDefinition.spec): unknown field "version" in io.k8s.apiextensions-apiserver.pkg.apis.apiextensions.v1.CustomResourceDefinitionSpec]
```

lspecian-olx:
> @lspecian-olx when i tried to install Operator on a cluster with `v1.19.9` version, got the below error. Are you also seeing same?
> 
> ```
> Error: unable to build kubernetes objects from release manifest: error validating "": error validating data: [ValidationError(CustomResourceDefinition.spec): unknown field "additionalPrinterColumns" in io.k8s.apiextensions-apiserver.pkg.apis.apiextensions.v1.CustomResourceDefinitionSpec, ValidationError(CustomResourceDefinition.spec): unknown field "subresources" in io.k8s.apiextensions-apiserver.pkg.apis.apiextensions.v1.CustomResourceDefinitionSpec, ValidationError(CustomResourceDefinition.spec): unknown field "validation" in io.k8s.apiextensions-apiserver.pkg.apis.apiextensions.v1.CustomResourceDefinitionSpec, ValidationError(CustomResourceDefinition.spec): unknown field "version" in io.k8s.apiextensions-apiserver.pkg.apis.apiextensions.v1.CustomResourceDefinitionSpec]
> ```

Yes, I'm also seeing the same, preparing a PR with a fix for the issue https://github.com/pravega/zookeeper-operator/issues/338

PR #297, opened 2021-01-25
Fixed zk image to run correctly on OpenShift

lamOrigin007:
### Change log description

* Changed location of `zu.jar` file in zookeeper image
* Fixed path to `zu.jar` file in shell scripts

### Purpose of the change

Correct launch of the zookeeper in OpenShift. Using the `/root/` folder in containers is bad practice.

### How to verify it

Tested in OpenShift version 4.5:
* install operator
* create ZookeeperCluster object

PR #334, opened 2021-05-26
Issue 333: Fix Pod SecurityContext config error in ZK charts

### LINKED ISSUE #333: [Bug]: Pod.securityContext setting in zookeeper chart values.yaml disappear in created zookeeper StatefulSet
### Description

Pod.securityContext setting in zookeeper chart values.yaml disappear in created zookeeper StatefulSet.

### Importance

must-have

### Location

https://github.com/gxglls/zookeeper-operator/blob/fd714b40e4148221dd9ab0c27b5693f51a95ee2a/pkg/zk/generators.go#L108
```golang
if reflect.DeepEqual(v1.PodSecurityContext{}, z.Spec.Pod.SecurityContext) { // should !reflect.DeepEqual() here
    podSpec.SecurityContext = z.Spec.Pod.SecurityContext	
}
```

Fixed by myself in related pr

gxglls:
### Change log description

1. Fix Pod SecurityContext config error in ZK charts

### Purpose of the change
Close #333 

### What the code does

1. Fix Pod SecurityContext config error in ZK charts

### How to verify it

Create a ZK CR with  Pod.SecurityContext  config, then check the created ZK StatufulSet config if contained 
correct Pod.SecurityContext

### DIFF
--- a/pkg/zk/generators.go
+++ b/pkg/zk/generators.go
@@ -175,7 +175,7 @@ func makeZkPodSpec(z *v1beta1.ZookeeperCluster, volumes []v1.Volume) v1.PodSpec 
 		Affinity:   z.Spec.Pod.Affinity,
 		Volumes:    append(z.Spec.Volumes, volumes...),
 	}
-	if reflect.DeepEqual(v1.PodSecurityContext{}, z.Spec.Pod.SecurityContext) {
+	if !reflect.DeepEqual(v1.PodSecurityContext{}, z.Spec.Pod.SecurityContext) {
 		podSpec.SecurityContext = z.Spec.Pod.SecurityContext
 	}
 	podSpec.NodeSelector = z.Spec.Pod.NodeSelector


anishakj:
@gxglls  have you verified zookeeper pods are coming up if security context is mentioned?  could you please mention what are the fields mentioned under that

gxglls:
> @gxglls have you verified zookeeper pods are coming up if security context is mentioned? could you please mention what are the fields mentioned under security context in the testing
I built zk operator docker image from my branch and tested it, SecurityContext is correctly attached to the generated zk sts and pods now, and zookeeper pods are coming up without error log. The value I tested in SecurityContext is `runAsUser` and `runAsGroup`, I think other SecurityContext's value will no problem as well as，because  It should be that this is not directly related to zk, only related to the pod strategy of k8s

gxglls:
> @gxglls have you verified zookeeper pods are coming up if security context is mentioned? could you please mention what are the fields mentioned under security context in the testing

I built zk operator docker image from my branch and tested it, SecurityContext is correctly attached to the generated zk sts and pods now, and zookeeper pods are coming up without error log. The value I tested in SecurityContext is runAsUser and runAsGroup, I think other SecurityContext's value will no problem as well as，because It should be that this is not directly related to zk, only related to the pod strategy of k8s

anishakj:
> > @gxglls have you verified zookeeper pods are coming up if security context is mentioned? could you please mention what are the fields mentioned under security context in the testing
> 
> @anishakj I built zk operator docker image from my branch and tested it, SecurityContext is correctly attached to the generated zk sts and pods now, and zookeeper pods are coming up without error log. The value I tested in SecurityContext is runAsUser and runAsGroup, I think other SecurityContext's value will no problem as well as，because It should not directly related to zk, only related to the pod strategy of k8s

When tested in my environment, have got this error in zk-0 pod and it  didnt come up 
`/usr/local/bin/zookeeperStart.sh: line 127: /data/zoo.cfg.dynamic: Permission denied
`
So felt we need to support initcontainers before adding this fix
Which environment have you tried

gxglls:
> When tested in my environment, have got this error in zk-0 pod and it didnt come up
> `/usr/local/bin/zookeeperStart.sh: line 127: /data/zoo.cfg.dynamic: Permission denied `
> So felt we need to support initcontainers before adding this fix
> Which environment have you tried
can you show what security context fields and value in your test

anishakj:
> > When tested in my environment, have got this error in zk-0 pod and it didnt come up
> > `/usr/local/bin/zookeeperStart.sh: line 127: /data/zoo.cfg.dynamic: Permission denied `
> > So felt we need to support initcontainers before adding this fix
> > Which environment have you tried
> 
> can you show me what security context fields and value in your test

``` securityContext:
    runAsGroup: 1001
    runAsUser: 1001
```

gxglls:
runAsUser set to 1001 will run zk pod fail as expected，because /data/zoo.cfg.dynamic file's owner is root which uid is 0
```
root@iomesh-zookeeper-0:/apache-zookeeper-3.5.9-bin# ls -l /data/zoo.cfg.dynamic
-rw-r--r--. 1 root root 113 May 27 02:23 /data/zoo.cfg.dynamic
```

In my case, I set runAsUser to 0. Because my environment is openshift, in openshift, container will run in an user which has an random uid such as 1000630000 for safety，but if zk pod run in uid 1000630000, it will run fail due to permission issues for read /root/zu.jar, like issue https://github.com/pravega/zookeeper-operator/pull/297 (This issue is merged？ okay). 

In k8s, container run in user root default, so zookeeper-operator can work well. 

So I think user run the container should be customize, and hhere is already a logic error in this line of code, user can't be customize

anishakj:
> runAsUser set to 1001 will run zk pod fail as expected，because /data/zoo.cfg.dynamic file's owner is root which uid is 0
> 
> ```
> root@iomesh-zookeeper-0:/apache-zookeeper-3.5.9-bin# ls -l /data/zoo.cfg.dynamic
> -rw-r--r--. 1 root root 113 May 27 02:23 /data/zoo.cfg.dynamic
> ```
> 
> In my case, I set runAsUser to 0. Because my environment is openshift, in openshift, container will run in an user which has an random uid such as 1000630000 for safety，but if zk pod run in uid 1000630000, it will run fail due to permission issues for read /root/zu.jar, like issue #297 (This issue is merged？ okay).
> 
> In k8s, container run in user root default, so zookeeper-operator can work well.
> 
> So I think user run the container should be customize, and hhere is already a logic error in this line of code, user can't be customize

Make sense, we can fix the permission issue in a separate PR

anishakj (approved):
LGTM

Issue #348, opened 2021-06-10
trigger cluster redeploy when static configs are modified


aparajita89:
### Description

when static configs are changed (such as initTime, etc which go in zoo.cfg), the zookeeper cluster needs to be restarted in order for the configs to take effect. ~the proposal is to add a flag in the helm chart values to optionally trigger a cluster redeploy when static configs are modified.~

[EDIT 10-Aug-21]: the finalized approach for cluster restart is as follows.
implementing a generic rolling restart feature via the CRD:
* introduce a new field "triggerRollingRestart" as part of the CRD spec
* user modifies the CRD to set "triggerRollingRestart" to true
* operator detects this change and restarts all the pods with the specs currently available in the CRD
* operator then sets the value of "triggerRollingRestart" to false and the flow completes
this approach would work irrespective of the deployment method chosen by users (helm/manual). it would also mean that the state of the cluster is still entirely controlled via the CRD.

### Importance

required feature.
in case of large zookeeper clusters this would become a required feature as currently there is no way to trigger a cluster redeploy via the operator. the only way is to manually delete a pod and to wait for the operator to bring back the pod with the new configs, and then to move on the the next pod.

### Location

~changes would be required in the helm chart template zookeeper.yaml and in the corresponding values.yaml files.~
[EDIT 10-Aug-21]: changes are needed in the operator's controller implementation and the CRD definition.

### Suggestions for an improvement

~a common way to mitigate this issue is to create an annotation which contains the checksum of the config map to monitor. when the config map changes, its checksum would also change and this new checksum needs to be updated in the annotation of the pod. the change in value of the annotation would trigger a recreation of the pod. this is described in detail [here](https://blog.questionable.services/article/kubernetes-deployments-configmap-change/).
this workflow can be automated easily using helm charts. the impact on users would be a simple change in values.yaml file.~
[EDIT 10-Aug-21]: when "triggerRollingRestart" is set to true in the CRD instance, operator will add an annotation to the cluster pods with the current time as the value. it then sets the value of "triggerRollingRestart" to false. this will trigger a one-time cluster restart.

aparajita89:
*bump*

anishakj:
@aparajita89 Currently, we are working on other priority items and will be working on it later. Would you like to contribute by raising a PR?

aparajita89:
yes, i can do that

aparajita89:
@anishakj i've raised the PR, can you check?

anishakj:
[...]

PR #349, opened 2021-06-17
issue #348 : added option to trigger cluster restart on static config change

aparajita89:
### Change log description

when static configs are changed (such as initTime, etc which go in zoo.cfg), the zookeeper cluster needs to be restarted in order for the configs to take effect. this change is to add a flag in the helm chart values to optionally trigger a cluster redeploy when static configs are modified.

### Purpose of the change

Refer https://github.com/pravega/zookeeper-operator/issues/348

### What the code does

these changes will compute a sha256 hash of the values which go into zoo.cfg via the config map. the hash is set as the value for a custom annotation in the pod spec for the zookeeper nodes. whenever the helm chart is uprgaded, the hash will be recomputed and the annotation will be updated. this update to the annotation will trigger a pod restart for all the nodes in the zookeeper cluster.

### How to verify it

1. create a zookeeper cluster using the helm chart
2. update the static configs in values.yaml for zookeeper chart
3. run "helm upgrade <chart-install-name> charts/zookeeper" from the root location of the project
4. check that the zookeeper pods are getting restarted by k8s using "kubectl get pods"

### DIFF
--- a/charts/zookeeper/README.md
+++ b/charts/zookeeper/README.md
@@ -54,6 +54,7 @@ The following table lists the configurable parameters of the zookeeper chart and
 | Parameter | Description | Default |
 | ----- | ----------- | ------ |
 | `replicas` | Expected size of the zookeeper cluster (valid range is from 1 to 7) | `3` |
+| `restartOnConfigChange` | Determines whether the zookeeper cluster needs to be automatically restarted when static configs present in zoo.cfg are changes | false |
 | `image.repository` | Image repository | `pravega/zookeeper` |
 | `image.tag` | Image tag | `0.2.11` |
 | `image.pullPolicy` | Image pull policy | `IfNotPresent` |
--- a/charts/zookeeper/templates/zookeeper.yaml
+++ b/charts/zookeeper/templates/zookeeper.yaml
@@ -82,9 +82,14 @@ spec:
     env:
 {{ toYaml .Values.pod.env | indent 6 }}
     {{- end }}
-    {{- if .Values.pod.annotations }}
+    {{- if or (.Values.pod.annotations) (eq .Values.restartOnConfigChange true) }}
     annotations:
+      {{- if eq .Values.restartOnConfigChange true }}
+      checksum/config: {{ .Values.config | toString | sha256sum }}
+      {{- end }}
+      {{- if .Values.pod.annotations }}
 {{ toYaml .Values.pod.annotations | indent 6 }}
+      {{- end }}
     {{- end }}
     {{- if .Values.pod.securityContext }}
     securityContext:
--- a/charts/zookeeper/values.yaml
+++ b/charts/zookeeper/values.yaml
@@ -1,4 +1,5 @@
 replicas: 3
+restartOnConfigChange: false
 
 image:
   repository: pravega/zookeeper
--- a/charts/zookeeper/values.yaml
+++ b/charts/zookeeper/values.yaml
@@ -1,5 +1,4 @@
 replicas: 3
-
 restartOnConfigChange: false
 
 image:
--- a/charts/zookeeper/values.yaml
+++ b/charts/zookeeper/values.yaml
@@ -1,4 +1,5 @@
 replicas: 3
+
 restartOnConfigChange: false
 
 image:


anishakj:
@aparajita89 In my testing, observed that `helm upgrade` is triggering restart of pods even when `restartOnConfigChange` is set to `false` and it is coming up with the new config values. But when editing zookeeper cluster using `kubectl edit zk zookeeper` and changing the config values, pods are not getting restarted automatically. Could you please verify this behaviour?

aparajita89:
this is happening because removal of the annotation is causing pod restart with the new values set in the config. i'm checking if adding a pre-install hook to the helm chart to get the old value of the annotation and setting that is going to help fix the issue.

anishakj:
> this is happening because removal of the annotation is causing pod restart with the new values set in the config. i'm checking if adding a pre-install hook to the helm chart to get the old value of the annotation and setting that is going to help fix the issue.

Also `kubectl edit zk zookeeper` and changing any of the config values is no triggering pod restart

aparajita89:
@anishakj in a cluster where the checksum/config value is already set in the annotations, if we try to update the configs by setting restartOnConfigChange to false, the value of the annotation becomes null and this triggers the pod restart. to avoid this, i tried to pull the existing value of the checksum/config annotation into a helm variable, and then to set the same value  to checksum/config if restartOnConfigChange is false but to compute the checksum again if it's true.
this approach did not work because even though we can get the existing checksum/config value using a pre install hook, it is not possible to set the value of a helm variable to this value, from within the helm template file.
our requirement is to trigger a cluster restart when the static configs change and we are ok with the cluster restart every time the config changes. if this would be useful for the wider community as well then i can update my PR to make the checksum computation mandatory and remove the restartOnConfigChange flag entirely.

aparajita89:
changing the config value in the CRD would not change the value of the annotation. we would need to make changes to the operator code so that it watches for changes in the static configs and updates the checksum/config annotation in the pod spec. k8s would then trigger the pod restart since the value of the annotation has now changed.

perhaps it would be more useful to abandon the idea of a cluster restart when static configs change, and to consider implementing a more generic rolling restart feature via the CRD.
how this would work:
* introduce a new field "triggerRollingRestart" as part of the CRD spec
* user modifies the CRD to set "triggerRollingRestart" to true
* operator detects this change and restarts all the pods with the specs currently available in the CRD
* operator then sets the value of "triggerRollingRestart" to false and the flow completes

this approach would work irrespective of the deployment method chosen by users (helm/manual). it would also mean that the state of the cluster is still entirely controlled via the CRD.

in case of a static config change, the CRD would be updated with the new static configs. this would not have any impact on the existing cluster. "triggerRollingRestart" can then be set to true to restart the pods on demand.

aparajita89:
> > this is happening because removal of the annotation is causing pod restart with the new values set in the config. i'm checking if adding a pre-install hook to the helm chart to get the old value of the annotation and setting that is going to help fix the issue.
> 
> Also `kubectl edit zk zookeeper` and changing any of the config values is not triggering pod restart

changing the config value in the CRD would not change the value of the annotation. we would need to make changes to the operator code so that it watches for changes in the static configs and updates the checksum/config annotation in the pod spec. k8s would then trigger the pod restart since the value of the annotation has now changed.

perhaps it would be more useful to abandon the idea of a cluster restart when static configs change, and to consider implementing a more generic rolling restart feature via the CRD.
how this would work:

introduce a new field "triggerRollingRestart" as part of the CRD spec
user modifies the CRD to set "triggerRollingRestart" to true
operator detects this change and restarts all the pods with the specs currently available in the CRD
operator then sets the value of "triggerRollingRestart" to false and the flow completes
this approach would work irrespective of the deployment method chosen by users (helm/manual). it would also mean that the state of the cluster is still entirely controlled via the CRD.

in case of a static config change, the CRD would be updated with the new static configs. this would not have any impact on the existing cluster. "triggerRollingRestart" can then be set to true to restart the pods on demand.

anishakj:
> > > this is happening because removal of the annotation is causing pod restart with the new values set in the config. i'm checking if adding a pre-install hook to the helm chart to get the old value of the annotation and setting that is going to help fix the issue.
> > 
> > 
> > Also `kubectl edit zk zookeeper` and changing any of the config values is not triggering pod restart
> 
> changing the config value in the CRD would not change the value of the annotation. we would need to make changes to the operator code so that it watches for changes in the static configs and updates the checksum/config annotation in the pod spec. k8s would then trigger the pod restart since the value of the annotation has now changed.
> 
> perhaps it would be more useful to abandon the idea of a cluster restart when static configs change, and to consider implementing a more generic rolling restart feature via the CRD.
> how this would work:
> 
> introduce a new field "triggerRollingRestart" as part of the CRD spec
> user modifies the CRD to set "triggerRollingRestart" to true
> operator detects this change and restarts all the pods with the specs currently available in the CRD
> operator then sets the value of "triggerRollingRestart" to false and the flow completes
> this approach would work irrespective of the deployment method chosen by users (helm/manual). it would also mean that the state of the cluster is still entirely controlled via the CRD.
> 
> in case of a static config change, the CRD would be updated with the new static configs. this would not have any impact on the existing cluster. "triggerRollingRestart" can then be set to true to restart the pods on demand.

Sure, This looks like a cleaner approach.

aparajita89:
will raise another PR for this

PR #360, opened 2021-07-02
ISSUE-356 Add support for more zoo.cfg configuration parameters

### LINKED ISSUE #356: Add support for more zoo.cfg configuration parameters
### Description

In our environment we are looking into tweaking `clientPortListenBacklog` and `serverCnxnFactory` zoo.cfg parameters to tune the zookeeper server(s) listening sockets.

### Importance

Should-have

we could extend the existing support to add these parameters.

amuraru:
### Change log description

Added a new `spec.config.additionalConfig` field to allov custom zoo.cfg parameters to passed to zookeeper

### Purpose of the change

Fixes #356


pkg/zk/generators.go
```diff
@@ -247,7 +247,8 @@ func MakeHeadlessService(z *v1beta1.ZookeeperCluster) *v1.Service {
 
 func makeZkConfigString(z *v1beta1.ZookeeperCluster) string {
 	ports := z.ZookeeperPorts()
-	return "4lw.commands.whitelist=cons, envi, conf, crst, srvr, stat, mntr, ruok\n" +
+
+	zkConfig := "4lw.commands.whitelist=cons, envi, conf, crst, srvr, stat, mntr, ruok\n" +
 		"dataDir=/data\n" +
 		"standaloneEnabled=false\n" +
 		"reconfigEnabled=true\n" +
```
anishakj, line 254: We  have to ensure that  these values are not overridden by the user,
```"dataDir=/data\n" +
		"standaloneEnabled=false\n" +
		"reconfigEnabled=true\n" +
                 "SkipACL=yes \n"
```
If we move these fields below, will that serve the purpose?

amuraru: good catch @anishakj !

Moved the optional parameters at the top of the file.
This way - even if they declare one of the reserved params, this being a java properties file - last entry wins

anishakj (changes requested):
Please update the helm charts as well. Also increase the test coverage and generate deep copy file

amuraru:
>Please update the helm charts as well. 

Good call - done


> Also increase the test coverage and generate deep copy file
Not sure I know how the deep copy is regenerated? Couldn't find a Makefile target. Can you expand on this please?

amuraru:
Getting the following error when trying to regenerate k8s files:

 ```
operator-sdk generate k8s
INFO[0000] Running deepcopy code-generation for Custom Resource group versions: [zookeeper:[v1beta1], ]
F0705 18:55:49.112842   30715 deepcopy.go:885] Hit an unsupported type invalid type for invalid type, from github.com/pravega/zookeeper-operator/pkg/apis/zookeeper/v1beta1.ClusterCondition

```

Any idea what am I doing wrong?

amuraru:
Addressed your review @anishakj  - please take another look

anishakj:
> Addressed your review @anishakj - please take another look

Sure, Thanks for the quick update

anishakj (approved):
LGTM

PR #363, opened 2021-07-06
Add generate Makefile target to generate operator-sdk resources

### LINKED ISSUE #364: Makefile is not providing a target to regenerate operator-sdk managed files
### Description

The operator is relying on (rather old) operator-sdk to (re)generate CRDs and types files.
For a contributor it's quite confusing on how this should be used.
As such - I propose we extend the Makefile and add a new target : `generate` that when called should regenerate all these files.

### Importance

SHOULD-HAVE to ease external contributions

### Location

Makefile

See Description

amuraru:
### Change log description

- Add a new `generate` Makefile target to generate k8s and CRD resources via operator-sdk
- Update dependencies:
  - Go 1.13 -> 1.16
  - operator-sdk: 0.17 -> 0.19
  - sigs.k8s.io/controller-runtime v0.5.2 -> v0.6.5
  - k8s v0.17.5 -> v0.19.13
 

Fixes #364


anishakj (changes requested):
Can we make 2 different versions of CRD, one for v1beta1 and other for v1

amuraru:
> Can we make 2 different versions of CRD, one for v1beta1 and other for v1. 

I couldn't find a way to do that using operator-sdk.
My preference would be to generate v1 raw version and keep the v1beta1 compat in helm chart only


>Also upgrade the kubernetes version here so that our e2e tests will pass https://github.com/pravega/zookeeper-operator/blob/master/.github/workflows/ci.yaml#L33

done, thanks for the pointer

anishakj:
> > Can we make 2 different versions of CRD, one for v1beta1 and other for v1.
> 
> I couldn't find a way to do that using operator-sdk.
> My preference would be to generate v1 raw version and keep the v1beta1 compat in helm chart only
> 
> > Also upgrade the kubernetes version here so that our e2e tests will pass https://github.com/pravega/zookeeper-operator/blob/master/.github/workflows/ci.yaml#L33
> 
> done, thanks for the pointer

If we didnt mention `operator-sdk generate crds`  crd version here `v1beta1` crd will get generated

amuraru:
> If we didnt mention crd version here in `operator-sdk generate crds` `v1beta1` crd will get generated

i know that but again do we want to stick with v1beta1 ? It's going EOL in couple of months once k8s 1.22 is released

cmd/manager/main.go
```diff
@@ -52,13 +52,6 @@ func printVersion() {
 
 func main() {
 	flag.Parse()
-
-	// The logger instantiated here can be changed to any logger
-	// implementing the logr.Logger interface. This logger will
-	// be propagated through the whole operator, generating
-	// uniform and structured logs.
-	logf.SetLogger(logf.ZapLogger(false))
-
```
anishakj, line 61: why are we removing this? will it cause any logging format change in operator logs?

amuraru: good catch - missed a local commit here

anishakj:
@amuraru Could you please verify operator upgrades are working fine?

charts/zookeeper-operator/templates/_helpers.tpl
```diff
@@ -3564,14 +3559,13 @@ openAPIV3Schema:
...
-                description: ContainerPort represents a network port in a single container.
+                description: ContainerPort represents a network port in a single
+                  container.
                 properties:
                   containerPort:
                     description: Number of port to expose on the pod's IP address.
                       This must be a valid port number, 0 < x < 65536.
                     format: int32
-                    minimum: 1
-                    maximum: 65535
```
anishakj, line 3574: some of  the openAPI Schema validations got removed. Could you please fix those?

anishakj (approved):
LGTM

PR #369, opened 2021-07-19
Add generate Makefile target to generate operator-sdk resources ADDENDUM

amuraru:
### Change log description

helm install fails to install the new CRD in k8s 1.15
Fixing by reverting to using helm conditional on these old versions


anishakj (approved):
LGTM

PR #370, opened 2021-07-19
Issue 363: Fixing CRD to work with k8 1.15 version

amuraru:
Fix issue in CRD

### Change log description

Change log description
helm install fails to install the new CRD in k8s 1.15
Fixing by reverting to using helm conditional on these old versions
### Purpose of the change

 Fixes #363

### What the code does

Fixed the CRD 

### How to verify it

Verified operator and cluster install works fine


anishakj (approved):
LGTM

PR #373, opened 2021-07-26
Issue #348: trigger cluster restart via crd field

### LINKED ISSUE #348: trigger cluster redeploy when static configs are modified

aparajita89:
### Change log description

implementing a generic rolling restart feature via the CRD
* introduce a new field "triggerRollingRestart" as part of the CRD spec
* user modifies the CRD to set "triggerRollingRestart" to true
* operator detects this change and restarts all the pods with the specs currently available in the CRD
* operator then sets the value of "triggerRollingRestart" to false and the flow completes
this approach would work irrespective of the deployment method chosen by users (helm/manual). it would also mean that the state of the cluster is still entirely controlled via the CRD.

### Purpose of the change
Fixes #348
implementing a generic rolling restart feature via the CRD

### What the code does

when "triggerRollingRestart" is set to true, operator will add an annotation to the cluster pods with the current time as the value. it then sets the value of "triggerRollingRestart" to false. this will trigger a one-time cluster restart.

### How to verify it

edit the crd instance to have "triggerRollingRestart" under the cluster spec set to true

### DIFF
--- a/deploy/crds/zookeeper.pravega.io_zookeeperclusters_crd.yaml
+++ b/deploy/crds/zookeeper.pravega.io_zookeeperclusters_crd.yaml
@@ -3659,6 +3659,9 @@ spec:
                   will be using It can take either Ephemeral or persistence Default
                   StorageType is Persistence storage
                 type: string
+              triggerRollingRestart:
+                description: if set to true, triggers a cluster restart. this value should be set to false by the operator once the restart completes for all pods.
+                type: boolean
               volumeMounts:
                 description: VolumeMounts defines to support customized volumeMounts
                 items:
--- a/charts/zookeeper-operator/Chart.yaml
+++ b/charts/zookeeper-operator/Chart.yaml
@@ -1,8 +1,8 @@
 apiVersion: v2
 name: zookeeper-operator
 description: Zookeeper Operator Helm chart for Kubernetes
-version: 0.2.11
-appVersion: 0.2.11
+version: 0.2.12
+appVersion: 0.2.12
 keywords:
 - zookeeper
 - storage
--- a/charts/zookeeper-operator/README.md
+++ b/charts/zookeeper-operator/README.md
@@ -42,7 +42,7 @@ The following table lists the configurable parameters of the zookeeper-operator 
 | Parameter | Description | Default |
 | ----- | ----------- | ------ |
 | `image.repository` | Image repository | `pravega/zookeeper-operator` |
-| `image.tag` | Image tag | `0.2.11` |
+| `image.tag` | Image tag | `0.2.12` |
 | `image.pullPolicy` | Image pull policy | `IfNotPresent` |
 | `crd.create` | Create zookeeper CRD | `true` |
 | `rbac.create` | Create RBAC resources | `true` |
--- a/charts/zookeeper-operator/templates/_crd_openapiv3schema.tpl
+++ b/charts/zookeeper-operator/templates/_crd_openapiv3schema.tpl
@@ -681,14 +681,13 @@ openAPIV3Schema:
                             type: string
                         required:
                         - containerPort
-                        {{- if semverCompare "< 1.18-0" .Capabilities.KubeVersion.GitVersion }}
-                        - protocol
-                        {{- end }}
                         type: object
                       type: array
                       x-kubernetes-list-map-keys:
                       - containerPort
+                      {{- if semverCompare "< 1.18-0" .Capabilities.KubeVersion.GitVersion }}
                       - protocol
+                      {{- end }}
                       x-kubernetes-list-type: map
                     readinessProbe:
                       description: 'Periodic probe of container service readiness.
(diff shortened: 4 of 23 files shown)


deploy/crds/zookeeper.pravega.io_zookeeperclusters_crd.yaml
```diff
@@ -3659,6 +3659,9 @@ spec:
                   will be using It can take either Ephemeral or persistence Default
                   StorageType is Persistence storage
                 type: string
+              triggerRollingRestart:
```
amuraru, line 3662: would it make sense to have a `revision` that can be an incremental version/sequence instead of a boolean switch?
This way the users can update it with a deployment timestamp if they want to kick a rolling restart.

aparajita89: if we change this to a revision number, then we will also need to support a rollback to a previous version. in this case, we will need to store the previous crd versions in order to be able to rollback to it. this is very similar to the functionality already provided by helm charts.
the deployment timestamp is already captured in the "restartTime" annotation which is being added to the pods so perhaps it would not be needed to capture it here as well. what do you think?

anishakj:
@aparajita89 Could you please add a functional test for triggering restart of pods, please refer here. (https://github.com/pravega/zookeeper-operator/tree/master/test/e2e)

Also the codecov calculates coverage under each package, so you have to add unit tests in `https://github.com/pravega/zookeeper-operator/blob/master/pkg/apis/zookeeper/v1beta1/zookeepercluster_types_test.go`  as well.

aparajita89:
sure, will check

aparajita89:
@anishakj  which version of operator-sdk are you using to run e2e scripts? i am not able to run them from the version installed via homebrew.

anishakj:
> @anishakj which version of operator-sdk are you using to run e2e scripts? i am not able to run them from the version installed via homebrew.

@aparajita89 we are on `v0.19.4`

aparajita89:
@anishakj just an update, the e2e test is succeeding on local but is failing on the automated build triggered by the git hook. i'm still trying to debug this. have you seen this happening before? perhaps i have missed out something in the test case.

anishakj:
> @anishakj just an update, the e2e test is succeeding on local but is failing on the automated build triggered by the git hook. i'm still trying to debug this. have you seen this happening before? perhaps i have missed out something in the test case.

@aparajita89 Could you please try removing these lines https://github.com/aparajita89/zookeeper-operator/blob/redeploy/test/e2e/rolling_restart_test.go#L43-#L47 ,so that it will take the cluster version as `0.2.12`. There were some issues in `0.2.9`  in minikube setup to bring the second zk pod.

aparajita89:
> > @anishakj just an update, the e2e test is succeeding on local but is failing on the automated build triggered by the git hook. i'm still trying to debug this. have you seen this happening before? perhaps i have missed out something in the test case.
> 
> @aparajita89 Could you please try removing these lines https://github.com/aparajita89/zookeeper-operator/blob/redeploy/test/e2e/rolling_restart_test.go#L43-#L47 ,so that it will take the cluster version as `0.2.12`. There were some issues in `0.2.9` in minikube setup to bring the second zk pod.

ok let me try

aparajita89:
@anishakj it worked! please check now.

anishakj (changes requested):
@aparajita89 Please update CRD in https://github.com/pravega/zookeeper-operator/blob/master/charts/zookeeper-operator/templates/_crd_openapiv3schema.tpl
Also add the new field `triggerRollingRestart` in  configuration of https://github.com/pravega/zookeeper-operator/blob/master/charts/zookeeper/README.md

aparajita89:
> @aparajita89 Please update CRD in https://github.com/pravega/zookeeper-operator/blob/master/charts/zookeeper-operator/templates/_crd_openapiv3schema.tpl
> Also add the new field `triggerRollingRestart` in configuration of https://github.com/pravega/zookeeper-operator/blob/master/charts/zookeeper/README.md

updated it

aparajita89:
@anishakj let me know if i need to make further changes

anishakj (approved):
LGTM

PR #383, opened 2021-08-24
Updating Zookeeper to 3.6.3 to bring in some upstream bug fixes

nikore:
### Change log description

Updating zookeeper docker image to 3.6.3 to fix some bugs, namely https://github.com/apache/zookeeper/pull/1630 which we have seen time to time in our environment. 

### Purpose of the change

Fixes ZOOKEEPER-4220

### How to verify it

We did a local `make build-zk-image` and deployed it to our clusters from our image repo. We have been running 3.6.3 for a few weeks without issue we updated to latest master yesterday + this fix and have been running it for 24 hours.

PR #384, opened 2021-08-25
Update imagepullsecrets in service account on existing clusters/serviceaccounts

nikore:
### Change log description

Updates the `imagePullSecrets` section of the `ServiceAccount` on reconciliation 

### What the code does

Noticed when working on #383 that if I updated the `imagePullSecrets` setting to use a private repo it wouldn't update the `ServiceAccount`. This meant I would have to either manually edit the `ServiceAccount` or delete the cluster and make a new one. 

This allows updating the `imagePullSecrets` section in the `ServiceAccount` on existing clusters. 

### How to verify it

Deploy a zookeeper cluster, then update `spec.Pod.ImagePullSecrets` with a new secret and wait for the cluster to reconcile. I also added a unit test covering this use case.


anishakj:
@nikore Please raise an issue corresponding to this. And looks like the newly added lines are not covered by unit tests. Could you please have a look?

PR #385, opened 2021-08-25
Prevent status updates from triggering reconcile loops

### LINKED ISSUE #389: Operator Reconciles endlessly
### Description

The reconcile loop is running endlessly often <1ms between each run per CRD. If you have a lot of CRDs it adds a lot of load to the kubeAPI server. On my cluster I was seeing >100 req/sec caused by this operator. This appears to be caused by the status update to the CRD as that does generate a new event but doesn't update the metadata nor object data. 

### Importance

must-have, if you have any large number of zookeeper clusters you would generate a ton of load on kubeapi server. In my current setup I would have to upscale the kubeapi server size to handle the load. 

My own my internal operators I have had to implement the GenrationChangedPredicate to stop status updates from triggering a loop. You can read about [GenerationChagnedPredicate](https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/predicate/predicate.go#L139) Operator-Framework has its own version here: https://github.com/operator-framework/operator-lib/blob/main/predicate/nogeneration.go#L24 depending on the version of operator-framework used in this project. 

In general predicates filter events before they are enqueued for processing; with out one you get all events regardless if they are ones you need to react to or not.

nikore:
### Change log description

Prevents the operator from reconciling every time it updates the status. 

### Purpose of the change

Fixes #389 

### What the code does

I noticed in the logs that the reconcile loop was running none stop, some times with only a few milisecond pauses. In my own operators I have had the same problem with updating the status object triggering a loop, if you use a predicate to filter events you can prevent that from happening. The build in [predicate.GenerationChangedPredicate{}](https://github.com/operator-framework/operator-sdk/blob/v0.19.4/pkg/predicate/predicate.go#L39) does a good job at preventing that and still allowing through other changes. 

### How to verify it

Start up the operator, create a cluster and it should reconcile once every 30seconds.

### DIFF
--- a/pkg/controller/zookeepercluster/zookeepercluster_controller.go
+++ b/pkg/controller/zookeepercluster/zookeepercluster_controller.go
@@ -16,6 +16,7 @@ import (
 	"strconv"
 	"time"
 
+	"github.com/operator-framework/operator-sdk/pkg/predicate"
 	"k8s.io/client-go/kubernetes/scheme"
 	"sigs.k8s.io/controller-runtime/pkg/client/fake"
 
@@ -71,7 +72,7 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
 
 	// Watch for changes to primary resource ZookeeperCluster
 	err = c.Watch(&source.Kind{Type: &zookeeperv1beta1.ZookeeperCluster{}},
-		&handler.EnqueueRequestForObject{})
+		&handler.EnqueueRequestForObject{}, predicate.GenerationChangedPredicate{})
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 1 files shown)


anishakj:
@nikore How can i validate the changes? Couldn't observe any thing different from the logs.

```
{"level":"info","ts":1630335567.1193426,"logger":"controller_zookeepercluster","msg":"Updating existing admin server service","Request.Namespace":"default","Request.Name":"zookeeper","Service.Namespace":"default","Service.Name":"zookeeper-admin-server"}
{"level":"info","ts":1630335567.1225502,"logger":"controller_zookeepercluster","msg":"Updating zookeeper status","Request.Namespace":"default","Request.Name":"zookeeper","StatefulSet.Namespace":"default","StatefulSet.Name":"zookeeper"}
````

nikore:
So you wont notice much as this just prevents the loop from running unless metadata or the actual object was updated. 

I discovered this by looking at the logs for the zookeeper-operator, and seeing how frequently `msg":"Updating zookeeper status"` showed up for a specific ZK cluster. With this change you should see it every 30 seconds, without this change you should see if much more frequently, in my env it was sub second. Granted I have a pretty limited test base but I saw the problem on k8s 1.18, 1.19 and 1.20 in GKE and EKS. I didn't test things like k3s, minikube, etc. Its possible the problem doesn't exist there.

anishakj (approved):
LGTM

PR #386, opened 2021-08-25
Add omitempty to optional probe fields

### LINKED ISSUE #378: Empty Probes in Spec do not play nice with the go-client validation
### Description

The Probes and all the fields within the probes (livenessProbe, readinessProbe, and all probe options), all have the `//+optional` tag added to make them optional in the CRD. However, when using the go client (from my experience writing unit tests for the solr-operator), I am still not allowed to submit a ZookeeperCluster with empty probes, or probes fields. 

### Importance

I can use the `zkCluster.WithDefaults()` method before submitting the object to the kube-client, however that really ties the Solr Operator down to making sure the running version of the Zookeeper Operator matches the version that the Solr Operator has in its dependencies. It's a little safer if the Solr Operator can just make changes to the fields it cares about, and let the Zookeeper Operator do the defaulting.

### Location

https://github.com/pravega/zookeeper-operator/blob/master/pkg/apis/zookeeper/v1beta1/zookeepercluster_types.go#L154  
https://github.com/pravega/zookeeper-operator/blob/master/pkg/apis/zookeeper/v1beta1/zookeepercluster_types.go#L157   

I think this can be fixed by merely adding `,omitempty` on the json part of the field variables.

For the [probe options](https://github.com/pravega/zookeeper-operator/blob/master/pkg/apis/zookeeper/v1beta1/zookeepercluster_types.go#L284), it might be better to make the minimums 1, add the `,omitempty` and keep the `// +optional` tag. That way it only prints out the values that have been provided. This is unrelated to the error I'm seeing, just a small improvement I think.

HoustonPutman:
### Change log description

_Fix issue with empty probes section in ZookeeperCluster using the golang client. Adds omitempty to the json serialization spec._

### Purpose of the change

Fixes #378


pkg/apis/zookeeper/v1beta1/zookeepercluster_types.go
```diff
@@ -155,14 +155,14 @@ type ZookeeperClusterSpec struct {
...
+	Probes *Probes `json:"probes,omitempty"`
 }
 
 type Probes struct {
 	// +optional
-	ReadinessProbe *Probe `json:"readinessProbe"`
+	ReadinessProbe *Probe `json:"readinessProbe,omitempty"`
 	// +optional
-	LivenessProbe *Probe `json:"livenessProbe"`
+	LivenessProbe *Probe `json:"livenessProbe,omitempty"`
```
anishakj, line 165: Do we need to add `omitempty`  for probe structure at line no:291-303  as well?

PR #399, opened 2021-10-14
Fixes issue#398 that decreasing replicas will make zookeeper unrecoverable when zookeeper not running.

### LINKED ISSUE #398: if zookeeper is not running, then decreasing  replicas will make cluster of zookeeper Unrecoverable
### Description

when a cluster of zookeeper is not running for some error, then decreasing  replicas will delete pod automatically.
pod exec zookeeperTeardown.sh  to connect zookeeper will fail ,then it will not remove node on this cluster of zookeeper.
But the pod have been delete without update zookeeper configure on zoo.cfg
### Importance

must-have

### Location

ZNODE_PATH="/zookeeper-operator/$CLUSTER_NAME"
CLUSTERSIZE=`java -Dlog4j.configuration=file:"$LOG4J_CONF" -jar /root/zu.jar sync $ZKURL $ZNODE_PATH`
echo "CLUSTER_SIZE=$CLUSTERSIZE, MyId=$MYID"
if [[ -n "$CLUSTERSIZE" && "$CLUSTERSIZE" -lt "$MYID" ]]; then
  # If ClusterSize < MyId, this server is being permanantly removed.
  echo $?
fi

fix on zookeepercluster_controller.go:reconcileStatefulSet
If ClusterSize decrease, do reconfig remove here

stop-coding:
### Change log description

Fixes the bug that decreasing replicas will make zookeeper unrecoverable when zookeeper not running.

### Purpose of the change

Fixes #398 

### What the code does

Add protection for setting Stateful when zookeeper not running.

### How to verify it

1. Create an cluster that size is 3 (kubectl create -f zk.yaml).
2. Wait all pod running, named: zk-0\zk-1\zk-2.
3. Delete zk-1\zk-2 pod, make cluster of zookeeper unable to provide services.
4. "kubectl edit zk" that change replicas to 1 immediately.
5. Wait some time, replicas will decrease to 1.
6. Now, checking zk-0 is all right?


pkg/controller/zookeepercluster/zookeepercluster_controller.go
```diff
@@ -616,7 +643,9 @@ func (r *ReconcileZookeeperCluster) reconcileClusterStatus(instance *zookeeperv1
 	instance.Status.Members.Unready = unreadyMembers
 
 	//If Cluster is in a ready state...
-	if instance.Spec.Replicas == instance.Status.ReadyReplicas && (!instance.Status.MetaRootCreated) {
+	// instance.Spec.Replicas is just an expected value that we set it, but it maybe not take effect by k8s.
+	// So we should check that instance.Status.Replicas is equal to ReadyReplicas, which means true true status of pods.
+	if instance.Status.Replicas == instance.Status.ReadyReplicas && (!instance.Status.MetaRootCreated) {
```
anishakj, line 648: If we wont compare with `spec.Replicas` status wont be shown correctly rt, in the case of scaling

stop-coding:
> @stop-coding have one more question. How you made the service of zookeeper cluster to stop? will it be possible to add a test for the same

@anishakj 
Delete more than half pods of zookeeper after editing the replicas on zk immediately.
It will also happen that some node which pod is running have network failure  when we edit the replicas.

PR #401, opened 2021-10-14
[Issue-400] add disableFinalizer flag

realAaronWu:
### Change log description

add DisableFinalizer config and skip adding zkFinalziers when set it to true.

### Purpose of the change

Issue-400

### What the code does

skip adding zkFinalziers when DisableFinalizer set to true.

### How to verify it

Set disableFinalizer to true in values and deploy zookeeper operator, the deployed zookeeper  clusters should have the finalizers as empty.And when delete operator before deleting zookeeper  clusters, zk cluster deletion will not be blocked.


anishakj (approved):
LGTM

PR #406, opened 2021-10-21
Issue#398 that decreasing replicas will make zookeeper unrecoverable when zookeeper not running.

### LINKED ISSUE #398: if zookeeper is not running, then decreasing  replicas will make cluster of zookeeper Unrecoverable

stop-coding:
Change log description
Fixes the bug that decreasing replicas will make zookeeper unrecoverable when zookeeper not running.

Purpose of the change
Fixes #398

What the code does
Add protection for setting Stateful when zookeeper not running.
If zookeeper not running, we will prohibited to update replicas status until zookeeper resume.
When user decrease replicas value, it will remove node with reconfig firstly.
Keep do that remove node with reconfig on preStop before pod exit.

How to verify it
Create an cluster that size is 3 (kubectl create -f zk.yaml).
Wait all pod running, named: zk-0\zk-1\zk-2.
Delete zk-1\zk-2 pod, it make cluster of zookeeper unable to provide services.
"kubectl edit zk" that change replicas to 1 immediately.
Wait some time, replicas will decrease to 1.
Now, checking that:
Is zk-0 is all right?

### DIFF
--- a/pkg/controller/zookeepercluster/zookeepercluster_controller.go
+++ b/pkg/controller/zookeepercluster/zookeepercluster_controller.go
@@ -14,6 +14,7 @@ import (
 	"context"
 	"fmt"
 	"strconv"
+	"strings"
 	"time"
 
 	"github.com/operator-framework/operator-sdk/pkg/predicate"
@@ -295,7 +296,22 @@ func (r *ReconcileZookeeperCluster) reconcileStatefulSet(instance *zookeeperv1be
 
 			data := "CLUSTER_SIZE=" + strconv.Itoa(int(newSTSSize))
 			r.log.Info("Updating Cluster Size.", "New Data:", data, "Version", version)
-			r.zkClient.UpdateNode(path, data, version)
+			err = r.zkClient.UpdateNode(path, data, version)
+			if err != nil {
+				return fmt.Errorf("Error updating cluster size %s: %v", path, err)
+			}
+			// #398 if decrease node, remove node immediately after update node successfully.
+			if newSTSSize < foundSTSSize {
+				var removes []string
+				for myid := newSTSSize + 1; myid <= foundSTSSize; myid++ {
+					removes = append(removes, strconv.Itoa(int(myid)))
+				}
+				r.log.Info("It will do reconfig to remove id:%s", strings.Join(removes, ","))
+				err := r.zkClient.ReconfigToRemove(removes)
+				if err != nil {
+					return fmt.Errorf("Error reconfig remove id:%s", strings.Join(removes, ","))
+				}
+			}
 		}
 		err = r.updateStatefulSet(instance, foundSts, sts)
 		if err != nil {
--- a/pkg/controller/zookeepercluster/zookeepercluster_controller.go
+++ b/pkg/controller/zookeepercluster/zookeepercluster_controller.go
@@ -280,6 +280,8 @@ func (r *ReconcileZookeeperCluster) reconcileStatefulSet(instance *zookeeperv1be
 		foundSTSSize := *foundSts.Spec.Replicas
 		newSTSSize := *sts.Spec.Replicas
 		if newSTSSize != foundSTSSize {
+			// If zookeeper is not running, it must stop update replicas.
+			// Until zookeeper is running and the client connect it successfully, decreasing Replicas will take effect.
 			zkUri := utils.GetZkServiceUri(instance)
 			err = r.zkClient.Connect(zkUri)
 			if err != nil {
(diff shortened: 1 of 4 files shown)


pkg/zk/zookeeper_client_test.go
```diff
@@ -53,5 +55,11 @@ var _ = Describe("Zookeeper Client", func() {
 		It("err5 should be not nil", func() {
 			Ω(err5).ShouldNot(BeNil())
 		})
+		It("err6 should be not nil", func() {
+			Ω(err6).ShouldNot(BeNil())
+		})
+		It("err should be not nil", func() {
```
jkhalack, line 61: Probably should be `"err6 should be not nil"` 🤔

pkg/controller/zookeepercluster/zookeepercluster_controller.go
```diff
@@ -295,7 +298,31 @@ func (r *ReconcileZookeeperCluster) reconcileStatefulSet(instance *zookeeperv1be
...
+				if err != nil {
+					return fmt.Errorf("Error GetConfig %v", err)
+				}
+				r.log.Info("Get zookeeper config.", "Config: ", config)
+				for myid := newSTSSize + 1; myid <= foundSTSSize; myid++ {
+					if strings.Contains(config, "server."+strconv.Itoa(int(myid))+"=") {
+						removes = append(removes, strconv.Itoa(int(myid)))
+					}
+				}
+				// The node that have been removed witch reconfig alse can still provide services for all online clients.
```
anishakj, line 318: could u please fix typo `false`

pkg/controller/zookeepercluster/zookeepercluster_controller.go
```diff
@@ -295,7 +298,31 @@ func (r *ReconcileZookeeperCluster) reconcileStatefulSet(instance *zookeeperv1be
...
+				}
+				r.log.Info("Get zookeeper config.", "Config: ", config)
+				for myid := newSTSSize + 1; myid <= foundSTSSize; myid++ {
+					if strings.Contains(config, "server."+strconv.Itoa(int(myid))+"=") {
+						removes = append(removes, strconv.Itoa(int(myid)))
+					}
+				}
+				// The node that have been removed with reconfig also can still provide services for all online clients.
+				// So We can remove it firstly, it will avoid to error that client can't connect to server on preStop.
+				r.log.Info("Do reconfig to remove node.", "Remove ids", strings.Join(removes, ","))
```
anishakj, line 320: we are doing the same operation of removing node in https://github.com/pravega/zookeeper-operator/blob/master/docker/bin/zookeeperTeardown.sh#L45. Do you still think removing node is required here?

jkhalack: @stop-coding Did you see `Do reconfig to remove node.` message being present in the logs in your use case?

I think @anishakj is suggesting that catching the `UpdateNode` error and returning on line 303 should be enough to fix the issue, hence lines 305 to 324 would never get executed.

stop-coding: @jkhalack  Sorry for my late reply.
Only  catching the  _UpdateNode error_ is not enough to ensure do reconfigure successfully on _preStop_, for example that pod exiting but the cluster broken again. We indeed hope that updating node size and doing reconfigure is atomicity, but it's not realistic. 

It known that updating "Spec.Replicas" to k8s will tell pod to create or exit. If it fail on scale down, we can stop updating "Spec.Replicas"  until cluster recovery, that will ensure have done reconfigure before pod exit.

So I think that doing reconfigure on checking the cluster scale down is better.

ranyhb:
i would like to know if this pull request will be included in the next release?

PR #412, opened 2021-11-19
Add missing namespaces to pre-delete-hooks.yaml in zookeeper-operator chart

### LINKED ISSUE #411: Namespaces are missing in pre-delete-hooks.yaml resources in the zookeeper-operator chart
### Description

If developers want to write programs for performing helm operations, e.g.`helm uninstall`, they would use the [`Init` function](https://github.com/helm/helm/blob/main/pkg/action/action.go#L365) to initialize the action configuration before calling [`Uninstall.Run`](https://github.com/helm/helm/blob/main/pkg/action/uninstall.go#L54), but this currently fails to uninstall the zookeeper-operator helm release since the `pre-delete-hooks` resources would be created in the `default` namespace and error `from server (Forbidden): zookeeperclusters.zookeeper.pravega.io is forbidden: User system:serviceaccount:default:zookeeper-operator-pre-delete" cannot list resource "zookeeperclusters" in API group` is seen in the pre-delete-hook pod because:
- The `Init` function didn't set namespace to the`KubeClient`
- The resources in [pre-delete-hooks.yaml](https://github.com/pravega/zookeeper-operator/blob/master/charts/zookeeper-operator/templates/pre-delete-hooks.yaml) don't have namespace set

> Note: using the CLI `helm uninstall` doesn't have the problem while uninstalling the zookeeper-operator helm chart
### Importance
Should-have

### Location
[pre-delete-hooks.yaml](https://github.com/pravega/zookeeper-operator/blob/master/charts/zookeeper-operator/templates/pre-delete-hooks.yaml)

Add the missing namespace to the pre-delete-hooks.yaml

panyuenlau:
Signed-off-by: Darren Lau <panyuenlau@gmail.com>

### Change log description

Add missing namespaces to the resources in pre-delete-hooks.yaml within the zookeeper-operator chart

### Purpose of the change

Fixes #411

### What the code does

This PR adds the `namespace` field to the resources in the pre-delete-hooks.yaml

### How to verify it

The programmatic way (as described in the issue) of uninstalling the zookeeper-operator would have all the pre-delete-hooks resources created in the namespace that the helm release uses, and the uninstallation works fine

### DIFF
--- a/charts/zookeeper-operator/templates/pre-delete-hooks.yaml
+++ b/charts/zookeeper-operator/templates/pre-delete-hooks.yaml
@@ -3,6 +3,7 @@ kind: ClusterRole
 apiVersion: rbac.authorization.k8s.io/v1
 metadata:
   name: {{ template "zookeeper-operator.fullname" . }}-pre-delete
+  namespace: {{ .Release.Namespace }}
   annotations:
     "helm.sh/hook": pre-delete
     "helm.sh/hook-weight": "1"
@@ -21,6 +22,7 @@ kind: ClusterRoleBinding
 apiVersion: rbac.authorization.k8s.io/v1
 metadata:
   name: {{ template "zookeeper-operator.fullname" . }}-pre-delete
+  namespace: {{ .Release.Namespace }}
   annotations:
     "helm.sh/hook": pre-delete
     "helm.sh/hook-weight": "1"
@@ -40,6 +42,7 @@ apiVersion: v1
 kind: ServiceAccount
 metadata:
   name: {{ template "zookeeper-operator.fullname" . }}-pre-delete
+  namespace: {{ .Release.Namespace }}
   annotations:
     "helm.sh/hook": pre-delete
     "helm.sh/hook-weight": "1"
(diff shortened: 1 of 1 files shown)


charts/zookeeper-operator/templates/pre-delete-hooks.yaml
```diff
@@ -3,6 +3,7 @@ kind: ClusterRole
 apiVersion: rbac.authorization.k8s.io/v1
 metadata:
   name: {{ template "zookeeper-operator.fullname" . }}-pre-delete
+  namespace: {{ .Release.Namespace }}
```
anishakj, line 6: namespace is not required for clusterrole/clusterrole binding, as it is not attached to namespace

anishakj (changes requested):

charts/zookeeper-operator/templates/pre-delete-hooks.yaml
panyuenlau: I see, thanks @anishakj! So we should also remove the namespace field from the clusterrole and clusterrolebinding within the `post-install-upgrade-hooks.yaml` as well?

anishakj: @panyuenlau  Not seeing `clusterRole` in `post-install-upgrade-hooks.yaml`

panyuenlau: @anishakj  oops my bad, I was looking into a wrong file

jkhalack (approved):
LGTM

anishakj (approved):
LGTM

PR #420, opened 2021-12-05
Upgrade to golang 1.17 and update dependencies

amuraru:
### Change log description

- Use golang 1.17
- Upgrade k8s deps to 1.20
 

### Purpose of the change

- keep deps up to date


anishakj:
@amuraru  Can we remove the `v1beta1` version of CRD from charts as well

amuraru:
I'd personally prefer to do that in a separate PR and not here. wdyt?

anishakj:
> I'd personally prefer to do that in a separate PR and not here. wdyt?

sure, that is absolutely fine

anishakj (approved):
LGTM

PR #437, opened 2022-02-22
Issue 432: zk operator pre delete hook fail in slow environments

### LINKED ISSUE #432: ZK-operator pre-delete hook fails in slow environments
### Description

Currently in the pre delete hook script in zookkeeper operator charts, the `activeDeadineSeconds` is being set to `20`. But In some environments the pre-delete hook takes more than 20 seconds to complete and because of this job fails . Moreover once the job is failed , the operator chart cannot be uninstalled normally.

### Importance

must-have

### Location

https://github.com/pravega/zookeeper-operator/blob/master/charts/zookeeper-operator/templates/pre-delete-hooks.yaml

 Need to remove `activeDeadlineSeconds` from the hook job spec to let the hook script run as long as needed and increase the `backoffLimit` say to 6  to allow for easier debugging of failed pods

nishant-yt:
### Change log description

Currently in the zookeeper-operator charts' pre delete hook-script, the `activeDeadlineSeconds` is set to 20 which causes job to fail in slow environments. Need to remove the `activeDeadlineSeconds` and increase the `backoffLimit`

### Purpose of the change

Fixes #432

### What the code does

Removes the `activeDeadlineSeconds` from the pre delete hook script of the zk-operator chart and add Security Context . Also bump the `backoffLimit` to 6 and update the `hooks.image.tag` in values.yaml file

### How to verify it

Verified by Installing the SDP on an Openshift cluster

### DIFF
--- a/charts/zookeeper-operator/templates/pre-delete-hooks.yaml
+++ b/charts/zookeeper-operator/templates/pre-delete-hooks.yaml
@@ -90,8 +90,7 @@ metadata:
     "helm.sh/hook-weight": "2"
     "helm.sh/hook-delete-policy": hook-succeeded, before-hook-creation, hook-failed
 spec:
-  backoffLimit: 1
-  activeDeadlineSeconds: 20
+  backoffLimit: 6
   template:
     metadata:
       name: {{ template "zookeeper-operator.fullname" . }}-pre-delete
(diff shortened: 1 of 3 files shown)


anishakj:
Closing this PR as it is taken care in PR https://github.com/pravega/zookeeper-operator/pull/441

PR #441, opened 2022-02-25
Issue 440: Added support to run zookeeper operator as non-root

### LINKED ISSUE #440: Provide support to run zookeeper-operator as non root
### Description

Currently , operator is running as root user

### Importance

 must-have
### Location

charts/zookeeper-operator/values.yaml

Provide support to configure security context to operator pod

anishakj:

### Change log description

Made changes in operator charts to configure security context.
Updated hook image to `1.23.2`
Added options in hooks to configure security context.
Active Delay seconds is removed in `pre-delete-hooks`

### Purpose of the change

 Fixes #440

### What the code does

Provide support  to run zookeeper-operator as non-root user

### How to verify it

Verify able to run zookeeper-operator as non root user

### DIFF
--- a/charts/zookeeper-operator/templates/post-install-upgrade-hooks.yaml
+++ b/charts/zookeeper-operator/templates/post-install-upgrade-hooks.yaml
@@ -94,6 +94,10 @@ spec:
       containers:
       - name: post-install-upgrade-job
         image: "{{ .Values.hooks.image.repository }}:{{ .Values.hooks.image.tag }}"
+        {{- if .Values.hooks.securityContext }}
+        securityContext:
+{{ toYaml .Values.hooks.securityContext | indent 10 }}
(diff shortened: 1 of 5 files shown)


jkhalack (approved):
LGTM

HoustonPutman:
Maybe I'm missing something here, but it doesn't look like the `{{Values.securityContext}}` is used anywhere in the zookeeper-operator deployment. The functionality was only added for the hooks...

janhoy:
I'm running into the same, trying to specify `securityContext` for the zookeeper-operator pod, which does not work. Is there a know workaround to this, and is there plans to add it in next release?

lfloris:
Has there been any progress on this issue at all? Running into the same issue where zookeeper operator pod does not run on restricted Kubernetes environments.

PR #449, opened 2022-03-29
Update go.mod with no replace directives & k8s ver

tmarcu:
The replace blocks should only be used temporarily unless an imported
package forces them on. To make zookeeper more importable by other
packages, don't use any replace directives that will require the parent
package to also set them.

Update Kubernetes version to 1.23.1 and controller runtime to 0.11.1.

Signed-off-by: Tudor Marcu <Tudor.Marcu@emc.com>

### Purpose of the change
Allow other packages that import zookeeper to more easily upgrade their overlapping k8s deps.

Fixes: https://github.com/pravega/zookeeper-operator/issues/450

### What the code does

N/A


go.mod
```diff
@@ -46,87 +40,32 @@ require (
...
 	gopkg.in/yaml.v3 v3.0.0-20210107192922-496545a6307b // indirect
-	k8s.io/apiextensions-apiserver v0.22.2 // indirect
-	k8s.io/klog/v2 v2.9.0 // indirect
-	k8s.io/kube-openapi v0.0.0-20211110013926-83f114cd0513 // indirect
-	k8s.io/utils v0.0.0-20210930125809-cb0fa318a74b // indirect
-	sigs.k8s.io/structured-merge-diff/v4 v4.1.2 // indirect
-	sigs.k8s.io/yaml v1.2.0 // indirect
-)
-
-replace (
```
amuraru, line 77: Please also remove https://github.com/pravega/zookeeper-operator/blob/610f8b213b9905be98997bf160d0d72187afe1a2/Makefile#L212-L213 if you think it's not needed anymore

anishakj:
@tmarcu  Could you please change clusterrole/clusterrole bindings in these files to `rbac.authorization.k8s.io/v1`.
config/rbac/all_ns_rbac.yaml
config/rbac/default_ns_rbac.yaml
test/e2e/resources/rbac-operator.yaml

go.mod
tmarcu: Good call thanks! :+1:

tmarcu:
> @tmarcu Changes are looking good. I have a query should we update the crds in charts as well?

ahh yes I think we should, we'll have to make sure the clusters currently running k8s < 1.23 accept the new CRDs and code.

anishakj (approved):
LGTM

PR #468, opened 2022-04-25
Issue 467: Fixing Leader election failure

### LINKED ISSUE #467: Leader election after node reboot fails
### Description

In VMware cluster, some pods are stuck in ProviderFailed state, and leader election function, provided by operator SDK, is unable to process that, so new pods are stuck in wait cycle.
Fix scans owners of leader lock config map, and deletes pods in ProviderFailed status.

### Importance

 must-have

### Location

main.go

Customise leader.Become function

anishakj:

### Change log description

use the customized leader.Become function as it handles certain pre-checks

### Purpose of the change

Fixes #467

### What the code does
When the operator-sdk version is upgraded, the customized leader function got replaced with operator-lib leader.Become function. So this PR reverts that and use customized function.

Also modifying operator.yaml to fix compatibility issues

### How to verify it

All e2e should pass


nishant-yt (approved):
LGTM

PR #469, opened 2022-04-28
Add hashed configs to pod annotations

### LINKED ISSUE #454: BUG: Zookeeper-operator does not restart pods when static configs are changed
### Description
When I change the fields under `spec.config`, zookeeper-operator does not issue a rolling update to roll out the changed config. For example, if I change the `spec.config.commitLogCount` field to `100`, the operator reconciles the configMap to reflect the change. So in the pod, `/conf/zoo.cfg` which is where the configMap mounted has the `commitLogCount` field set to `100`. But the `/data/conf/zoo.cfg` which is the config used by zookeeper still has the `commitLogCount` set to default value as `500`.

#### Steps to reproduce:
1. Deploy a simple zookeeper cluster with the following yaml:
```yaml
apiVersion: "zookeeper.pravega.io/v1beta1"
kind: "ZookeeperCluster"
metadata:
  name: "zookeeper"
spec:
  replicas: 3
```
2. Change the zookeeper yaml to configure the commitLogCount to 100
```yaml
apiVersion: "zookeeper.pravega.io/v1beta1"
kind: "ZookeeperCluster"
metadata:
  name: "zookeeper"
spec:
  replicas: 3
  config:
    commitLogCount: 100
```
3. Observe that the pods didn't restart, and go into the pod and check the config used by zookeeper in `/data/conf/zoo.cfg`:
<pre>
metricsProvider.exportJvmInfo=true
dataDir=/data
4lw.commands.whitelist=cons, envi, conf, crst, srvr, stat, mntr, ruok
syncLimit=2
<b>commitLogCount=500</b>
metricsProvider.httpPort=7000
snapSizeLimitInKb=4194304
standaloneEnabled=false
metricsProvider.className=org.apache.zookeeper.metrics.prometheus.PrometheusMetricsProvider
initLimit=10
minSessionTimeout=4000
snapCount=10000
admin.serverPort=8080
autopurge.purgeInterval=1
maxSessionTimeout=40000
maxCnxns=0
globalOutstandingLimit=1000
reconfigEnabled=true
skipACL=yes
autopurge.snapRetainCount=3
tickTime=2000
quorumListenOnAllIPs=false
preAllocSize=65536
maxClientCnxns=60
dynamicConfigFile=/data/conf/zoo.cfg.dynamic.100000010
root@zookeeper-0:/data/conf# cat zoo.cfg | grep con
4lw.commands.whitelist=cons, envi, conf, crst, srvr, stat, mntr, ruok
reconfigEnabled=true
dynamicConfigFile=/data/conf/zoo.cfg.dynamic.100000010
</pre>

### Importance

should-have

### Location

Zookeeper-operator is missing the functionality to restart the pods when config is changed.

We suggest to attach the hash of config as annotations to zookeeper's statefulSet's template. So that when the config is changed, the changed annotation would trigger statefulSet's rolling update.

hoyhbx:
### Change log description

- Automatically trigger statefulSet rolling update when config is changed
- Add hashed zookeeper configuration as the pod annotation in the statefulSet

### Purpose of the change

Fixes #454 

### What the code does

Add hashed zookeeper configuration as the pod annotation in the statefulSet.
The makeZkPodAnnotations function computes the hash for the zookeeper config string and calls mergeAnnotations to merge with the annotations from the CR.

There are several possible improvements regarding the patch:
1. I was not sure whether to start the function name with capital case or lower case (e.g. makeZkPodAnnotations or MakeZkPodAnnotations), since both code styles exist in the code base.
2. The `mergeAnnotations` function encodes the same logic as the `mergeLabels` function, we could reduce these two into a general `mergeMap` function to reduce the code redundency.
3. I am not sure if the changes in the `env.sh` could reflect to the zookeeper cluster without a restart. If that also requires a restart, then we can add the hash of result of `makeZkEnvConfigString` to the annotation map too.

### How to verify it

Similar to the steps mentioned in #454 
1. Deploy a default zookeeper cluster
2. Change one of the fields in the config field, and observe that the statefulSet performed a rolling update, and the config used by the zookeeper cluster is updated.


anishakj:
@hoyhbx  Could you please add an end-to-end test to validate this?

anishakj:
> ### Change log description
> * Automatically trigger statefulSet rolling update when config is changed
> * Add hashed zookeeper configuration as the pod annotation in the statefulSet
> 
> ### Purpose of the change
> Fixes #454
> 
> ### What the code does
> Add hashed zookeeper configuration as the pod annotation in the statefulSet. The makeZkPodAnnotations function computes the hash for the zookeeper config string and calls mergeAnnotations to merge with the annotations from the CR.
> 
> There are several possible improvements regarding the patch:
> 
> 1. I was not sure whether to start the function name with capital case or lower case (e.g. makeZkPodAnnotations or MakeZkPodAnnotations), since both code styles exist in the code base.
> 2. The `mergeAnnotations` function encodes the same logic as the `mergeLabels` function, we could reduce these two into a general `mergeMap` function to reduce the code redundency.
> 3. I am not sure if the changes in the `env.sh` could reflect to the zookeeper cluster without a restart. If that also requires a restart, then we can add the hash of result of `makeZkEnvConfigString` to the annotation map too.
> 
> ### How to verify it
> Similar to the steps mentioned in #454
> 
> 1. Deploy a default zookeeper cluster
> 2. Change one of the fields in the config field, and observe that the statefulSet performed a rolling update, and the config used by the zookeeper cluster is updated.

For the functions that are local name is starting with small letter
It is better to combine `mergeAnootations` and `mergeLabels`

PR #483, opened 2022-06-08
Issue 482: Reevaluating number of client connections dynamically in the teardown script

### LINKED ISSUE #482: Number of client connections not reevaluated dynamically in the teardown script
### Description

The `CONN_COUNT` variable is only evaluated once in the zookeeperTeardown.sh script (https://github.com/pravega/zookeeper-operator/blob/master/docker/bin/zookeeperTeardown.sh#L23-L32)
As a result, the loop would always take 30 seconds if there were client connections present when the script started to run, and ZK pod would terminate with `137` error code with the default termination grace period set to 30s (the same as in https://github.com/pravega/zookeeper-operator/issues/91)

### Importance

The failure makes it impossible to replace a failed node on K8s cluster (the affected ZK pod would not get shut down properly, hence not possible to migrate it).

### Location

https://github.com/pravega/zookeeper-operator/blob/master/docker/bin/zookeeperTeardown.sh#L23-L32

Move line 23 (evaluation of `CONN_COUNT`) inside the `for` loop, so it gets reevaluated every cycle and allows the code to break out of the loop earlier (then we might also have less time spent in upgrades, as in https://github.com/pravega/zookeeper-operator/issues/206).

nishant-yt:
Signed-off-by: Nishant Gupta <Nishant_Gupta3@dell.com>

### Change log description
Currently the CONN_COUNT variable is only evaluated once in the zookeeperTeardown.sh script 
As a result, the loop would always take 30 seconds if there were client connections present when the script started to run, and ZK pod would terminate with 137 error code with the default termination grace period set to 30s.  Need to make sure CONN_COUNT gets reevaluated every cycle and allows the code to break out of the loop earlier.

### Purpose of the change

Fixes #482

### What the code does

Moves the CONN_COUNT variable inside the for loop

### How to verify it

All e2e should pass

### DIFF
--- a/docker/bin/zookeeperTeardown.sh
+++ b/docker/bin/zookeeperTeardown.sh
@@ -20,8 +20,8 @@ LOG4J_CONF=/conf/log4j-quiet.properties
 
 # Wait for client connections to drain. Kubernetes will wait until the confiugred
 # "terminationGracePeriodSeconds" before focibly killing the container
-CONN_COUNT=`echo cons | nc localhost 2181 | grep -v "^$" |grep -v "/127.0.0.1:" | wc -l`
 for (( i = 0; i < 6; i++ )); do
+  CONN_COUNT=`echo cons | nc localhost 2181 | grep -v "^$" |grep -v "/127.0.0.1:" | wc -l`
   if [[ "$CONN_COUNT" -gt 0 ]]; then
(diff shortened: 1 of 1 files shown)


anishakj (approved):
LGTM

jkhalack (approved):
LGTM

PR #490, opened 2022-07-12
fix: Use HELM 3 best practices for CRDs

ryanfaircloth:
### Change log description

Current logic uses helm2 style conditions for CRDs which may or may not be installed prior to use when included in a subchart or as part of CD. Simply removing the if condition from the CRD sync step in the make file allows the helm3 supported method of opt out ``--skip-crds``

### Purpose of the change

Implement CRD best practices for helm3

### What the code does

Remove if/else logic as this is now built into helm3

### How to verify it

On a new cluster manually install CRDs then install chart using --skip-crds flag (confirm working)
On a new cluster do not manuall install CRDs then install without flag and confirm working


ryanfaircloth:
Note: Codecov is reporting a change in a file no present in the branch

amuraru (approved):

PR #515, opened 2022-11-23
Upgrade to zk 3.8, upgrade to temurin java-11, delete apt cache

### LINKED ISSUE #514: Release newer zookeeper docker image
### Description

The [pravega/zookeeper](https://hub.docker.com/r/pravega/zookeeper) docker image is 7 months old, and quite out of date on all levels, both Linux, Java (openjdk is no longer maintained) and Zookeeper (3.7.1 vs 3.8.0).

### Importance

High importance due to security vulnerabilities in existing image.

### Location

docker/Dockerfile

* Upgrade FROM tags

janhoy:
Closes #514

### Change log description

Upgrade to zookeeper 3.8. Upgrade to Temurin java-11. Delete apt cache

### Purpose of the change

See #514 

### What the code does

Builds a better docker image

### How to verify it

Build image, verify that it uses newer Zookeeper (3.8), newer Java release (Eclipse Temurin Openjdk-11), newer Linux release (Ubuntu Focal)


docker/Dockerfile
```diff
@@ -9,16 +9,17 @@
...
 ARG DOCKER_REGISTRY
-FROM  ${DOCKER_REGISTRY:+$DOCKER_REGISTRY/}openjdk:11-jdk
+FROM  ${DOCKER_REGISTRY:+$DOCKER_REGISTRY/}eclipse-temurin/11-jdk-focal
 RUN mkdir /zu
 COPY zu /zu
 WORKDIR /zu
 RUN ./gradlew --console=verbose --info shadowJar
 
-FROM ${DOCKER_REGISTRY:+$DOCKER_REGISTRY/}zookeeper:3.7.1
+FROM ${DOCKER_REGISTRY:+$DOCKER_REGISTRY/}zookeeper:3.8-temurin
```
janhoy, line 18: You may prefer to pin an exact version of the image?

anishakj:
@janhoy Could you please try scaling down of zookeeper cluster with the new image and see if dynamic config is generated. we have seen issue with `3.8.0` image. [ZOOKEEPER-4530](https://issues.apache.org/jira/browse/ZOOKEEPER-4530)

janhoy:
> @janhoy Could you please try scaling down of zookeeper cluster with the new image and see if dynamic config is generated. we have seen issue with `3.8.0` image. [ZOOKEEPER-4530](https://issues.apache.org/jira/browse/ZOOKEEPER-4530)

I'm not able to test this now, but the zk JIRA is not closed, so why would it work? Better try to get traction on that JIRA to find the root cause? I'm afraid I don't have spare cycles to help with that.

janhoy:
So what can we do to move this forward?

janhoy:
I believe we should still publish a docker image for ZK 3.8, even if we believe ZK to have a bug with dynamic config after down-scaling. Downscaling is not a very common operation, and we should still give users of zookeeper-operator access to 3.8 features / bugfixes. Then if anyone feels down-scaling of a cluster is a blocker for them, they can override tag to continue using 3.7.x or they can help fix the root cause.

I'm going to update my PR branch to 3.8.1

janhoy:
For those who want to test the image produced by this dockerfile, here is one I built on my account:

```bash
docker run --rm cominvent/pravega-zookeeper:3.8.1-temurin
```

https://hub.docker.com/layers/cominvent/pravega-zookeeper/3.8.1-temurin/images/sha256-0daecbc2e69dc9840bf3c6ae1361826675af2ef097e2e2f346a0b09306697c2c?context=explore

I had an issue building the amd64-image from my M1 mac with `docker buildx`, it was hanging during `gradlew`, so I had to pre-build the jar-file and just copy it when making this test repo.

aignatov (approved):
Good change. Checked ZK running from this Dockerfile. It works well

alewis001:
Hi. What needs to happen to have this change merged and a new release to include it? With the various managed Kubernetes offerings (EKS, AKS, LKE, etc.) they are all moving to a version of Kubernetes that has switched to cgroups v2 (EKS in 1.26, AKS in 1.25, etc.). That change causes Java below the latest builds of Java 8 and Java 11 to misread container CPU and memory. Java 15+ already supports cgroups v2. With this change to the Temurin based image of 11, this build would include 11.0.16 which has the cgroups v2 backported fixes/changes.

Without this change, users will see odd memory/CPU behaviour. (I.e. Java thinks it has the node's memoery, not what's allocated to the container).

diegs:
@alewis001 (and others who end up here), fwiw openjdk 11 should also have cgroups v2 after 11.0.16 (https://bugs.openjdk.org/browse/JDK-8283559).

i haven't verified but the openjdk base image has 11.0.16 (https://hub.docker.com/layers/library/openjdk/11-jdk/images/sha256-e81b7f317654b0f26d3993e014b04bcb29250339b11b9de41e130feecd4cd43c?context=explore), and a new image was built in #586 

still, it would be nice to switch to a more modern image. the openjdk base image is officially deprecated per the above link

alewis001:
> @alewis001 (and others who end up here), fwiw openjdk 11 should also have cgroups v2 after 11.0.16 (https://bugs.openjdk.org/browse/JDK-8283559).
> 
> i haven't verified but the openjdk base image has 11.0.16 (https://hub.docker.com/layers/library/openjdk/11-jdk/images/sha256-e81b7f317654b0f26d3993e014b04bcb29250339b11b9de41e130feecd4cd43c?context=explore), and a new image was built in #586
> 
> still, it would be nice to switch to a more modern image. the openjdk base image is officially deprecated per the above link

Hi @diegs, thanks for the response. As it was a long time ago I can't quite remember what we did but I think we were able to bump the 11 version on what we deploy.

Issue #518, opened 2022-12-07
solr-operator-zookeeper-operator-pre-delete gets scheduled on windows node


arogic:
### Description

We have an AKS Kubernetes Cluster with Linux and Windows Node Pools. We deploy the zookeeper-operator helm chart through the solr-operator helm chart. When we delete the solr-operator using helm,  the "solr-operator-zookeeper-operator-pre-delete" job gets scheduled on the Windows NodePool which obviously fails. Is there a way to pass NodeSelector values over for the "solr-operator-zookeeper-operator-pre-delete" job to be scheduled on the Linux Node Pool?

### Suggestions for an improvement

Option in the helm chart values file to define nodeSelector for the "solr-operator-zookeeper-operator-pre-delete" job.

Thanks in advance.

arogic:
I got around the issue by setting node taints on my AKS Windows Node Pool.

` 
"node_taints": [
    "app.kubernetes.io/name=solr:NoSchedule"
 ]
`

However it would be great, if NodeSelector would be carried over from the values file to the job.

PR #520, opened 2022-12-13
Adds nodeSelector, affinity and tolerations to the pre-delete hook Job

georgekaz:
make delete Job respect scheduling choices

### Change log description

Makes the delete job respect node scheduling the same as the post install hooks do. Without this, it fails in a mixed windows/linux cluster.

### Purpose of the change

Fixes https://github.com/pravega/zookeeper-operator/issues/518
Fixes https://github.com/apache/solr-operator/issues/382

### What the code does

Adds a node selector to control pre-delete Job scheduling

### How to verify it

Deploy and remove. See that it schedules where you expect it to.


jkhalack (approved):
One day we might want to split the hook and the operator settings (in case they would need to be different).

The current change LGTM, as it goes along with the other hook.

anishakj (approved):
LGTM

Issue #428, opened 2022-01-24
policy/v1beta1 PodDisruptionBudget is deprecated in v1.21+


cbuto:
### Description

The `PodDisruptionBudget` created for zookeeper clusters needs to be updated to use `policy/v1`.

```
W0124 15:31:48.254050       1 warnings.go:67] policy/v1beta1 PodDisruptionBudget is deprecated in v1.21+, unavailable in v1.25+; use policy/v1 PodDisruptionBudget
```

### Importance

This is fairly low importance right now but should be addressed before 1.25+ (which is a ways out)

### Location

https://github.com/pravega/zookeeper-operator/blob/master/pkg/controller/zookeepercluster/zookeepercluster_controller.go#L529
https://github.com/pravega/zookeeper-operator/blob/3359647e5044ed1e9918e84eae425136328b7709/pkg/zk/generators.go#L355

### Suggestions for an improvement

I'm not sure what the Kubernetes version support looks like for this project, but `policy/v1` is only available in 1.21+ clusters so that's something to consider.

cbuto:
thanks @anishakj! 

Is there a release cadence or do you have an idea of when the next release would be?

cbuto:
👋 friendly ping @anishakj about the above ^

anishakj:
@TeddyAndrieux we will do it by end of this month.

TeddyAndrieux:
Perfect, thank you :+1:

TomBillietKlarrio:
We'd also appreciate a release since we're working to upstep to 1.25 :pray: 
Thanks upfront!

PR #523, opened 2023-01-05
feat: Replace deprecated policy/v1beta1 PDB API

cbuto:
### Change log description
Replaces the deprecated PodDistruptionBudget policy/v1beta1 API with the new policy/v1 API.

### Purpose of the change

Closes https://github.com/pravega/zookeeper-operator/issues/428

### What the code does

Replaces the policy/v1beta1 API by simply changing the API version, there are no schema changes to the API, only a behavioral change - see https://kubernetes.io/docs/reference/using-api/deprecation-guide/#poddisruptionbudget-v125.

### How to verify it

A simple verification would be to manually create a ZK cluster on a 1.25 Kubernetes cluster and verify that the PDB created is using the new API. The e2e tests seem to cover this though so those seem to be sufficient.

### DIFF
--- a/controllers/zookeepercluster_controller.go
+++ b/controllers/zookeepercluster_controller.go
@@ -15,19 +15,19 @@ import (
 	"strconv"
 	"time"
 
-	"github.com/pravega/zookeeper-operator/pkg/controller/config"
-	"github.com/pravega/zookeeper-operator/pkg/utils"
-	"github.com/pravega/zookeeper-operator/pkg/yamlexporter"
-	"github.com/pravega/zookeeper-operator/pkg/zk"
 	"k8s.io/client-go/kubernetes/scheme"
 	"sigs.k8s.io/controller-runtime/pkg/client/fake"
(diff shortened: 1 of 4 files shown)


cbuto:
@anishakj would you be able to take a look at this PR in order to support Kubernetes 1.25+?

anishakj:
@cbuto Thanks for the PR. Could you please verify that operator upgrade is working fine?

cbuto:
> @cbuto Thanks for the PR. Could you please verify that operator upgrade is working fine?

@anishakj yep! I tested a simple upgrade path and everything worked fine. Since there isn't any difference between the APIs schemas, the `policy/v1beta1` are actually served as `policy/v1` and the only thing that changes is the `apiVersion`. 

One more thing related to upgrades is the `policy/v1` API is only available in Kubernetes 1.21+ so that will be the min version support.

jkhalack (approved):
LGTM

anishakj (approved):
LGTM

PR #526, opened 2023-02-05
Check orphan PVC before updating statefulSet

### LINKED ISSUE #513: Zookeeper pod keeps crashing when scaling down and up
### Description

_(Describe the feature, bug, question, proposal that you are requesting)_
We found that zookeeper pod keeps crashing after a scale-down workload and a scale-up workload.
We found the crash is caused by the undeleted PVC during the scale-down process. However, this problem still persists even when we specify the `reclaimPolicy: Delete`.

The root cause of this issue is because there is no guard for scale-up before finishing deleting the PVC. We found that zookeeper-operator checks if the statefulSet is under a ready state before proceeding with next reconcilation action, but this check misses checking for the undeleted OrphanPVCs.

### Importance

_(Indicate the importance of this issue to you (blocker, must-have, should-have, nice-to-have))_
Importance: must-have

### Location

The PVC cleanup code is here: https://github.com/pravega/zookeeper-operator/blob/c29d475794226ce38de021d72430ccde75a38d2a/controllers/zookeepercluster_controller.go#L707
The logic to upgrade the statefulSet is here: https://github.com/pravega/zookeeper-operator/blob/c29d475794226ce38de021d72430ccde75a38d2a/controllers/zookeepercluster_controller.go#L264

And we think that the operator should check if all OrphanPVCs are deleted before proceeding to the next upgrade.

hoyhbx:
### Change log description

- Check orphan PVCs are deleted before updating the statefulSet

### Purpose of the change

Fixes #513 

### What the code does

_(Detailed description of the code changes)_
If the ZooKeeper cluster is using Persistence storage and the reclaimPolicy is set to Delete, this change makes sure that the number of PVCs is not larger than the statefulSet replicas before updating the statefulSet.

### How to verify it

The bug #513 cannot be reproduced after this commit. E2E test is added to prevent regression


anishakj:
@hoyhbx Could you please increase the code coverage. Also the DCO check is failing

hoyhbx:
I will try to write a system test to cover these two branches

hoyhbx:
@anishakj, I wrote an e2e test to reproduce the issue.
During the process, I realize the previous fix does not work. I want to understand the code better before implementing the fix.

I want to confirm my understand for the usage of `ZooKeeper.Status.Replicas` and `ZooKeeper.Status.ReadyReplicas` before I implement the fix. The two fields seem to be used to reflect the StatefulSet's `Status.Replicas` and `Status.ReadyReplicas` fields, but they are only updated in this function: https://github.com/pravega/zookeeper-operator/blob/28d1f6917d2284c5673503816e110cf25555bd49/controllers/zookeepercluster_controller.go#L251

I want to implement a fix to make the `reconcileStatefulSet` function to return early if there are still orphan PVCs, by comparing the `ZooKeeper.Spec.Replicas` to `ZooKeeper.Status.Replicas`.  
The orphanPVC cleanup should be triggered by  comparing against `ZooKeeper.Status.Replicas` instead of `ZooKeeper.Spec.Replicas` here: https://github.com/pravega/zookeeper-operator/blob/28d1f6917d2284c5673503816e110cf25555bd49/controllers/zookeepercluster_controller.go#L717

But this doesn't work if the `ZooKeeper.Status.Replicas` is updated at the end of the `reconcileStatefulSet` function, because it would be skipped along with the StatefulSet update. I propose to update the `ZooKeeper.Status.Replicas` and `ZooKeeper.Status.Replicas` every reconcilation cycle, by either moving it earlier in the `reconcileStatefulSet` function, or move it to the `reconcileClusterStatus` function.

hoyhbx:
Please check the latest commit for the tentative fix. I moved the  `ZooKeeper.Status.Replicas` and `ZooKeeper.Status.ReadyReplicas` update to the `reconcileClusterStatus`.
For cleaning up orphanPVC, the fix gets the up to date StatefulSet to check for the need of cleaning up. It checks for the `STS.Spec.Replicas` against PVC count, because I found that in some corner cases `STS.Status.Replicas` may not up to date, causing it to delete the first PVC when starting the StatefulSet initially.

The e2e test added can reproduce the issue without the patch, and it passes after the patch

controllers/zookeepercluster_controller.go
```diff
@@ -707,21 +728,38 @@ func (r *ZookeeperClusterReconciler) getPVCCount(instance *zookeeperv1beta1.Zook
...
+			"PVC Count", pvcCount,
+			"Replicas Count", foundSts.Spec.Replicas)
+		if pvcCount > int(*foundSts.Spec.Replicas) {
 			pvcList, err := r.getPVCList(instance)
 			if err != nil {
 				return err
 			}
 			for _, pvcItem := range pvcList.Items {
 				// delete only Orphan PVCs
-				if utils.IsPVCOrphan(pvcItem.Name, instance.Spec.Replicas) {
```
anishakj, line 724: @hoyhbx why are we deleting pvcs based on Sts replicas. Operator is looking for zookeeper cluster resource. Is there any issue are you seeing if we delete based on instance.Spec.Replicas?

hoyhbx: This is to handle the race condition, where the operator has not yet deleted the orphan pvcs after scaling down, but the user scale the cluster back up. In that case, if `instance.Spec.Replicas` is used to delete old pvcs, the old pvcs will never get deleted.

Just as the added e2e test, when scaling down from 3 to 1, two pods are deleted, and statefulSet's replica gets down to 1. Then it takes sometime for the operator to delete the orphan pvcs. But before operator is able to delete the orphan pvcs, user scales up from 1 to 3, changing the `instance.Spec.Replicas` to 3. Then the old PVCs will never be deleted.

hoyhbx: Hi @anishakj , does the above explanation make sense to you?
The problem is basically because there is a race condition when deleting the PVC and user upscaling

anishakj:
@hoyhbx Could you please increase the code coverage for this?

hoyhbx:
Hi @anishakj , have you gotten a chance to look at the changes? We are happy to make improvements if you have some suggestions.

anishakj:
> Hi @anishakj , have you gotten a chance to look at the changes? We are happy to make improvements if you have some suggestions.

will perform some more tests from my side and let you know

hoyhbx:
Thanks @anishakj ! Just let us know if there is anything we could improve. We are also more than happy to fix the issue in the zookeeperStart.sh mentioned here: https://github.com/pravega/zookeeper-operator/issues/513#issuecomment-1502246369

hoyhbx:
Hi @anishakj , did you encounter any problem when testing this PR? Is there anything we can help?

iampranabroy:
@anishakj - Is this part of the release [v0.2.15](https://github.com/pravega/zookeeper-operator/releases/tag/v0.2.15). I don't see that as part of the change log. If not, any plans to include it in the new release? Thanks

anishakj:
> @anishakj - Is this part of the release [v0.2.15](https://github.com/pravega/zookeeper-operator/releases/tag/v0.2.15). I don't see that as part of the change log. If not, any plans to include it in the new release? Thanks

@hoyhbx Could you please increase the coverage

PR #530, opened 2023-02-14
Issue #410: fix reconcile if foreground cascading deletion is used

### LINKED ISSUE #410: [BUG] Unexpected resource creation when the ZookeeperCluster is already deleted
### Description

We find that the zookeeper operator, after deleting the PVC and removing the finalizer for a ZookeeperCluster CR with a non-zero DeletionTimestamp (in reconcileFinalizers), will continue running the reconciliation logic if there is no error:
```
func (r *ReconcileZookeeperCluster) reconcileFinalizers(instance *zookeeperv1beta1.ZookeeperCluster) (err error) {
	...
	if instance.DeletionTimestamp.IsZero() {
		...
	} else {
		if utils.ContainsString(instance.ObjectMeta.Finalizers, utils.ZkFinalizer) {
			if err = r.cleanUpAllPVCs(instance); err != nil {
				return err
			}
			instance.ObjectMeta.Finalizers = utils.RemoveString(instance.ObjectMeta.Finalizers, utils.ZkFinalizer)
			if err = r.client.Update(context.TODO(), instance); err != nil {
				return err
			}
		}
	}
	return nil
}
```
This can lead to unexpected resource creation, because after removing the finalizer of the CR, the CR is going to be deleted. When the CR is deleted, the resources (e.g., sts, service, secret, etc.) that are owned by the CR will also be deleted. If the operator does not immediately return after removing the finalizer, the original reconcile logic will find these resources do not exist and try to create these resources. For example, in reconcileStatefulSet the operator will create the zookeeper statefulset. Such creation is unexpected and unnecessary, as they will be deleted again since the owner CR does not exist.

### Importance

should-have

### Location

`reconcileFinalizers`:
https://github.com/srteam2020/zookeeper-operator/blob/a99c93ba06fbece2c262f7744517c2e537491c29/pkg/controller/zookeepercluster/zookeepercluster_controller.go#L664

Instead of creating the resources that are destined to be deleted immediately, we should make the operator return from reconcile if the pvc is deleted successfully and the finalizer is removed from the CR successfully, as no further actions should be taken if the CR is deleted.

We are willing to send a PR to help fix it.

lipov3cz3k:
### Change log description
Fix unexpected resource creation when the ZookeeperCluster is already deleted 

### Purpose of the change
Fixes #410 

### What the code does
Exit reconcile when resource has finalizer and is marked for delete. That's prevent to apply other reconcile functions to create already deleted resources.

### How to verify it

Delete `zookeepercluster` resource in `foreground` mode -> without this patch it goes to loop (you can fix it by deleting zookeeper in default `background` mode)
```bash
kubectl delete zookeeperclusters.zookeeper.pravega.io zookeeper  --cascade=foreground
```


janhoy:
I also experienced this multiple times. Have to explicitly use non-cascading delete for it to work.
Can someone please review and approve this?

janhoy:
@lipov3cz3k can you perhaps fix the two failed tests "DCO" and "codecov/patch", so it is more likely that the maintainers can approve and merge this?

anishakj:
@lipov3cz3k Could you please increase the code coverage?

controllers/zookeepercluster_controller.go
```diff
@@ -100,8 +100,11 @@ func (r *ZookeeperClusterReconciler) Reconcile(_ context.Context, request ctrl.R
 		}
 		return reconcile.Result{Requeue: true}, nil
 	}
+	if finalized, err := r.reconcileFinalizers(instance); err != nil || finalized {
```
jkhalack, line 103: It would be much cleaner if we returned an error when there is a need to do another attempt to run a reconcile loop.

controllers/zookeepercluster_controller.go
```diff
@@ -671,30 +674,30 @@ func (r *ZookeeperClusterReconciler) yamlConfigMap(instance *zookeeperv1beta1.Zo
...
 			}
 			instance.ObjectMeta.Finalizers = utils.RemoveString(instance.ObjectMeta.Finalizers, utils.ZkFinalizer)
 			if err = r.Client.Update(context.TODO(), instance); err != nil {
-				return err
+				return false, err
 			}
 		}
 	}
-	return nil
+	return true, nil
```
jkhalack, line 700: Would be cleaner to return some kind of error at this point instead of introducing a boolean.

jkhalack (changes requested):

anishakj:
@lipov3cz3k  Is it possible to increase the coverage?

Issue #508, opened 2022-10-28
Critical CVE-2022-37434 in zookeeper-operator image


AKamyshnikova:
### Description

CVE checker in docker images reported Found CVE in zookeeper-operator image version 0.2.14
https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2022-37434

### Importance

_must-have_

### Suggestions for an improvement

Build new image using latest updates.

PR #533, opened 2023-02-24
Use distroless image for operator image

AKamyshnikova:
To avoid security issues switch to usage of distroless image.
Also bumped go version to 1.19

Fixes: #508


Dockerfile
```diff
@@ -1,6 +1,6 @@
 ARG DOCKER_REGISTRY
 ARG ALPINE_VERSION=3.15
```
aignatov, line 2: I would change alpine version here also to the latest one 3.17

AKamyshnikova: Done

aignatov (approved):

aignatov (approved):

Dockerfile
```diff
@@ -28,14 +28,10 @@ COPY controllers/ controllers/
...
 
-FROM ${DOCKER_REGISTRY:+$DOCKER_REGISTRY/}alpine:${ALPINE_VERSION} AS final
-
+FROM gcr.io/distroless/static:nonroot AS final
 
 ARG PROJECT_NAME=zookeeper-operator
 
 COPY --from=go-builder /src/${PROJECT_NAME} /usr/local/bin/${PROJECT_NAME}
 
-RUN adduser -D ${PROJECT_NAME}
```
anishakj, line 38: @AKamyshnikova why are we removing the user

AKamyshnikova: @anishakj For distrolless image we cannot run adduser command. I will add update this.

AKamyshnikova: @anishakj Seems this setting is no longer required.

Dockerfile
```diff
@@ -28,14 +28,10 @@ COPY controllers/ controllers/
 RUN GOOS=linux GOARCH=amd64 CGO_ENABLED=0 go build -o /src/${PROJECT_NAME} \
     -ldflags "-X ${REPO_PATH}/pkg/version.Version=${VERSION} -X ${REPO_PATH}/pkg/version.GitSHA=${GIT_SHA}" main.go
 
-FROM ${DOCKER_REGISTRY:+$DOCKER_REGISTRY/}alpine:${ALPINE_VERSION} AS final
-
+FROM gcr.io/distroless/static:nonroot AS final
```
derekm, line 31: Currently available images are: https://github.com/GoogleContainerTools/distroless#what-images-are-available

That is, maybe we should use `gcr.io/distroless/static-debian11` directly instead of `gcr.io/distroless/static`. I think `static` is a link to the latest build `static-debian11`, but we should use the latest build explicitly, so that we have a reproducible build and so we don't auto-upgrade images when the `static` link changes.

derekm (changes requested):

Dockerfile
AKamyshnikova: Done

Dockerfile
```diff
@@ -28,14 +28,10 @@ COPY controllers/ controllers/
 RUN GOOS=linux GOARCH=amd64 CGO_ENABLED=0 go build -o /src/${PROJECT_NAME} \
     -ldflags "-X ${REPO_PATH}/pkg/version.Version=${VERSION} -X ${REPO_PATH}/pkg/version.GitSHA=${GIT_SHA}" main.go
 
-FROM ${DOCKER_REGISTRY:+$DOCKER_REGISTRY/}alpine:${ALPINE_VERSION} AS final
-
+FROM gcr.io/distroless/static-debian11:nonroot AS final
```
jkhalack, line 31: Please keep the option to use `DOCKER_REGISTRY` arg

jkhalack (changes requested):

Dockerfile
AKamyshnikova: DOCKER_REGISTRY is different for golang build image and distroless final. I will introduce DISTROLESS_DOCKER_REGISTRY.

Makefile
```diff
@@ -23,6 +23,7 @@ GIT_SHA=$(shell git rev-parse --short HEAD)
 TEST_IMAGE=$(TEST_REPO)-testimages:$(VERSION)
 DOCKER_TEST_PASS=testzkop@123
 DOCKER_TEST_USER=testzkop
+DISTROLESS_DOCKER_REGISTRY=gcr.io
```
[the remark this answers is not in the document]
AKamyshnikova: Yes, it can be dropped - I forgot to drop after I add default in Dockerfile

anishakj (approved):
LGTM

derekm (approved):
Looks good!

It might make sense to extract a distroless version like there is an alpine version, but this also looks great as-is.

PR #564, opened 2023-07-03
the default snapCount configuration doesn't mismatch the comment.

sangheee:
### Change log description

The default value of `snapCount` that the comment says is 100,000. 
But the actual value is set at 10,000.
I corrected this mismatch between them.

### Purpose of the change

None


anishakj:
@sangheee Here the comment needs to be corrected, default snapshot  count is `10000` https://zookeeper.apache.org/doc/r3.1.2/zookeeperAdmin.html

PR #576, opened 2023-11-20
Issue 572: Enable external option to client service

### LINKED ISSUE #572: Zk clientservice cannot be declared external
The operator creates 3 services: adminServerService, clientservice and headless. Out of which, only adminServerService could be declared as external (available outside of the cluster via nodeport).
This creates a problem when deploying an app that uses the zk as state manager because other apps may require access to that state. 

For example, Solr using the operator, zk operator is used to create a quorum that keeps tab on Solr's nodes. In order to write data to Solr one must contact the quorum (done through client port) and obtain a list of available nodes. Without this ability, Solr is practically unavailable.

This problem could be solved relatively simply by adding the option to make clientservice external to the operator

idolaman:
### Change log description

Adds the option to use external ZooKeeper in `ClientService`

### Purpose of the change

Fixes #572 

### What the code does

The code uses the same logic as AdminServerService boolean external field. in ClientService, to expose client-service outside 
the cluster, using LoadBalancer service type.

### How to verify it

System tests must pass, when clientService.external is true, client service in kubernetes type needs to be LoadBalancer (this test added).


almogtavor:
@anishakj can you check this out? This is a quite needed feature

Issue #581, opened 2023-12-08
Helm chart deployment does not expose metrics for collection


joshsouza:
### Description

At the present time, the Helm deployment of `zookeeper-operator` does not expose the `metrics` port to the cluster for collection by Prometheus or other tools. While metrics are available from `localhost` in the pod on port `6000`, they are bound to `127.0.0.1:6000` by default, and thus cannot be accessed by other pods in the cluster.

The Helm chart does not have any variables established to configure this at this time, which means that outside of modifications to the deployment outside of Helm, the metrics cannot be reached.

### Importance

Should-have: The metrics are generated, but not accessible, as was surely intended

### Location

`charts/solr-operator` (deployment/values)

### Suggestions for an improvement

I will provide a PR to go with this issue for my proposed fix, but anything that will make these metrics available by default or configurable to become available would be sufficient.

PR #582, opened 2023-12-08
Issue-581 Make metrics exposable based on helm variables

joshsouza:
Resolves: https://github.com/pravega/zookeeper-operator/issues/581


jkhalack:
Has there been any testing done to prove that the modified chart allows exposing metrics for Prometheus?

charts/zookeeper-operator/templates/operator.yaml
```diff
@@ -33,12 +33,13 @@ spec:
...
         imagePullPolicy: {{ .Values.image.pullPolicy }}
         ports:
-        - containerPort: 6000
+        - containerPort: {{ int .Values.metricsPort }}
           name: metrics
         command:
         - zookeeper-operator
-        {{- if .Values.disableFinalizer }}
         args:
+        - -metrics-bind-address={{ .Values.metricsBindAddress }}:{{ int .Values.metricsPort }}
```
jkhalack, line 41: Does it have to be `-metrics-bind-address` or `--metrics-bind-address`?

joshsouza: Based on: https://github.com/pravega/zookeeper-operator/blob/074d8b09facefb1ec7de7c0ed9142f44cd3d51d8/main.go#L65
And flag's documentation: https://pkg.go.dev/flag#hdr-Command_line_flag_syntax
My understanding is that it can be either. If there is a preference for double dashes, that's easy to adjust, but as `-disableFinalizer` below is a single dash, I went with single dash for consistency.

joshsouza:
> Has there been any testing done to prove that the modified chart allows exposing metrics for Prometheus?

I utilized helm template when writing the code to ensure the appropriate adjustments to the k8s resources, and deploying it to a local kind cluster before/after with a second `curl` pod deployed to test connectivity validates that without these changes I am unable to `curl <pod_ip>:6000/metrics` (connection refused) beforehand, and with `--set metricsBindAddress=0.0.0.0` in my helm update/install parameters, I am able to successfully reach the metrics endpoint from my `curl` pod afterwards.

While I didn't explicitly test that Prometheus itself was working, this should be a sufficient test to verify that the changes are functioning as intended, and should allow metrics gathering by other pods in the cluster.

jkhalack (approved):

anishakj (approved):
LGTM

PR #591, opened 2024-01-22
Issue 589: Add imagePullSecrets for Helm hooks

### LINKED ISSUE #589: Add imagePullSecrets to Helm Hooks in Zookeeper_Operator
### Description

Currently the service accounts/Job definitions used in the of Helm hooks doesn't have a way to incorporate the imagepull secrets.

### Importance

MustHave

### Location

https://github.com/pravega/zookeeper-operator/blob/master/charts/zookeeper-operator/templates/pre-delete-hooks.yaml#L39-L47
https://github.com/pravega/zookeeper-operator/blob/master/charts/zookeeper-operator/templates/pre-delete-hooks.yaml#L39-L47

Will try to add the global imagepullsecrets section under serviceaccount.

unguiculus:
### Change log description

* Adds imagePullSecrets to the service accounts for the Helm hooks
* Follows the same principle as the service account for the operator taking global imagePullSecrets into account

### Purpose of the change

Fixes #589

### What the code does

Adds imagePullSecrets to the YAML for the ServiceAccounts for the Helm hooks. This is important when images are pulled from a private registry, e.g. an internal proxy, such as Artifactory, or when using custom images.

### How to verify it

Render the templates using the following command:

```console
helm template zookeeper charts/zookeeper-operator \
    --show-only templates/post-install-upgrade-hooks.yaml \
    --show-only templates/pre-delete-hooks.yaml \
    --set hooks.serviceAccount.imagePullSecrets={'my-pull-secret'}
```

### DIFF
--- a/charts/zookeeper-operator/values.yaml
+++ b/charts/zookeeper-operator/values.yaml
@@ -57,6 +57,10 @@ tolerations: []
 annotations: {}
 
 hooks:
+  ## Optionally specify an array of imagePullSecrets. Will override the global parameter if set
+  serviceAccount:
+    imagePullSecrets: []
+
   backoffLimit: 10
   image:
     repository: lachlanevenson/k8s-kubectl
--- a/charts/zookeeper-operator/templates/post-install-upgrade-hooks.yaml
+++ b/charts/zookeeper-operator/templates/post-install-upgrade-hooks.yaml
@@ -51,6 +51,12 @@ metadata:
     "helm.sh/hook": post-install, post-upgrade
     "helm.sh/hook-weight": "1"
     "helm.sh/hook-delete-policy": hook-succeeded, before-hook-creation, hook-failed
+{{- if or .Values.global.imagePullSecrets .Values.hooks.serviceAccount.imagePullSecrets }}
+imagePullSecrets:
+{{- range (default .Values.global.imagePullSecrets .Values.hooks.serviceAccount.imagePullSecrets) }}
+  - name: {{ . }}
+{{- end }}
+{{- end }}
 
 ---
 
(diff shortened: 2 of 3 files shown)


krishnadas-menon:
@jkhalack Would you be able to have a look at this change? It would be really appreciated to get this change added to the helm charts. TIA

charts/zookeeper-operator/values.yaml
```diff
@@ -57,6 +57,10 @@ tolerations: []
 annotations: {}
 
 hooks:
+  ## Optionally specify an array of imagePullSecrets. Will override the global parameter if set
```
jkhalack, line 60: Please update the comment to explain that the array here is expected to contain secret names only, as we seem to have two different schemas already in the code:
- for `pod.imagePullSecrets` we are expecting key-value pair objects
https://github.com/pravega/zookeeper-operator/blob/master/charts/zookeeper/templates/zookeeper.yaml#L110
- for `serviceAccount.imagePullSecrets` we are expecting secret names
https://github.com/pravega/zookeeper-operator/blob/master/charts/zookeeper-operator/templates/service_account.yaml#L7
Otherwise I think it is good to go, unless you would think it necessary to update the schema for `pod.imagePullSecrets` as well (to be able to consume `global.imagePullSecrets` there).
@anishakj would you agree that updating `pod.imagePullSecrets` schema would be appropriate?

unguiculus: I tried to be consistent and matched what is already done in the `values.yaml`. See https://github.com/pravega/zookeeper-operator/blob/master/charts/zookeeper-operator/values.yaml#L30. If you like, I could update the whole chart in order to improve this.

anishakj (approved):
LGTM

krishnadasm:
@anishakj Could you please help me with the timeline for this change to be released. Appreciate if you could expedite the same. Thanks in advance.

crazyzhou:
Merged as admin.

PR #624, opened 2025-03-09
Issue 623: add security context override to operator deployment (#623)

### LINKED ISSUE #623: [Helm] securityContext is not implemented for operator
Currently, the deployment for the operator pod lacks a way to override the securityContext.

In the default values.yaml file, there is a `securityContext` section suggesting that it is implemented, but it is not made use of in the operator.yaml file.

The ability to override the securityContext is important for clusters where Pod Security Admission is enforced.

Hexoplon:
### Change log description

* Add podSecurityContext and securityContext overrides to Zookeeper Operator deployment in helm chart
* Updated README for helm chart

### Purpose of the change

Fixes #623 

### What the code does

Updated the Zookeper Operator helm chart to include the already existing securityContext section in the helm chart values.yaml file. Now, the securityContext values are correctly added to the operator Deployment. This ensures that the operator can run in a Kubernetes cluster where Pod Security Admission is enforced, and running in restricted mode.

### How to verify it

Add custom settings to securityContext and/or podSecurityContext, and verify that the deployment object has been updated using: `kubectl -n <namespace> describe deployment zookeeper-operator`

### DIFF
--- a/charts/zookeeper-operator/README.md
+++ b/charts/zookeeper-operator/README.md
@@ -58,8 +58,10 @@ The following table lists the configurable parameters of the zookeeper-operator 
 | `image.tag` | Image tag | `0.2.15` |
 | `labels` | Operator pod labels | `{}` |
 | `nodeSelector` | Map of key-value pairs to be present as labels in the node in which the pod should run | `{}` |
+| `podSecurityContext` | Security context for the pod | `{}` |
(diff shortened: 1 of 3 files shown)


Saremox:
We're making usage of `pod-security.kubernetes.io/enforce = restricted` in all our clusters as a hardening measure. Would love to see this PR in the next version.

